PR #5, opened 2023-01-06
Fix IUCN status for cockroaches

pjeweb:
The cockroaches seem not to have actually been evaluated yet.


abdullahmorrison:
You're right, thank you!

PR #13, opened 2023-01-13
Scroll to active animal on selection

### LINKED ISSUE #9: Feature: Scroll ambassadors list to active ambassador
It would be nice to scroll the list to the active ambassador when chosen by chat. I can send a pull request if you want.

I would probably would render some attribute (`data-ambassador-id={animal.name}`) to each button and then use something like:
```typescript
const top = animalList.current.querySelector(`[data-ambassador-id="${activeAnimal.name}"]`).offsetTop;
animalList.current.scrollTo({top, behavior: 'smooth'});
```
in the effect.

pjeweb:
Resolves #9

Tested in current Chrome, FF and Safari

### DIFF
--- a/src/pages/overlay/components/AmbassadorList.tsx
+++ b/src/pages/overlay/components/AmbassadorList.tsx
@@ -11,10 +11,29 @@ import AnimalButton from '../../../utils/global/animalButton/AnimalButton'
 import styles from './ambassadorList.module.css'
 import arrow from '../../../assets/arrow.jpg'
 
+const SCROLL_OFFSET = 200;
+
 export interface AmbassadorListProps{
     showAnimalList: boolean
     chatChosenAmbassador?: string
 }
+
+function scrollListToAnimal(listElement: HTMLDivElement | null, name: string) {
+    if (!listElement) {
+        return;
+    }
+
+    const anchorElement = listElement.querySelector(
+        `[data-animal-name="${name}"]`
+    );
+    if (anchorElement instanceof HTMLDivElement) {
+        listElement.scrollTo({
+            top: Math.max(0, anchorElement.offsetTop - SCROLL_OFFSET),
+            behavior: "smooth",
+        });
+    }
+}
+
 export default function AmbassadorList(props: AmbassadorListProps){
     const [animals] = useState(AnimalData)
     const [activeAnimal, setActiveAnimal] = useState<AnimalCardProps["cardData"] | null>()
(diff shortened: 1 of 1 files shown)


src/pages/overlay/components/AmbassadorList.tsx
```diff
@@ -11,10 +11,29 @@ import AnimalButton from '../../../utils/global/animalButton/AnimalButton'
 import styles from './ambassadorList.module.css'
 import arrow from '../../../assets/arrow.jpg'
 
+const SCROLL_OFFSET = 200;
```
abdullahmorrison, line 14: put variable inside scrollListToAnimal function. it is only used in that function.

src/pages/overlay/components/AmbassadorList.tsx
```diff
@@ -11,10 +11,29 @@ import AnimalButton from '../../../utils/global/animalButton/AnimalButton'
...
+
 export interface AmbassadorListProps{
     showAnimalList: boolean
     chatChosenAmbassador?: string
 }
+
+function scrollListToAnimal(listElement: HTMLDivElement | null, name: string) {
```
abdullahmorrison, line 16: Move function to the inside of AmbassadorList function. Move to under the useEffect

abdullahmorrison, line 16: convert function to arrow function

src/pages/overlay/components/AmbassadorList.tsx
```diff
@@ -34,6 +29,11 @@ function scrollListToAnimal(listElement: HTMLDivElement | null, name: string) {
     }
 }
 
+export interface AmbassadorListProps{
+    showAnimalList: boolean
+    chatChosenAmbassador?: string
+}
+
```
abdullahmorrison, line 36: remove new line in between AmbassadorListProps and AmbassadorList

abdullahmorrison (changes requested):

src/pages/overlay/components/AmbassadorList.tsx
pjeweb: should be fixed

pjeweb: should be fixed

src/pages/overlay/components/AmbassadorList.tsx
pjeweb: should be fixed

pjeweb:
Tried to match your code style better in general, like no semicolons.

abdullahmorrison (approved):

PR #26, opened 2023-01-22
Improve overlay scaling

### LINKED ISSUE #24: Improve AmbassadorList responsiveness
Could we make the list a bit taller in smaller viewports? According to the twitch dev docs [1] the overlay can be up to 5rem from top/bottom. Another idea would be to move the scroll arrow button in front of the list to gain a few more pixels. I can make a pull request if you would consider it.

Current:
Overlay at 1024 pixel width

Proposed arrow button above list:
Mockup arrow button above list

------------------

1) https://dev.twitch.tv/docs/extensions/designing/#video-overlay-extensions

pjeweb:
My first attempt on the improved scaling.

- Instead of scaling everything with `scale()` i opted to scale the list components separately with some `calc(x * clamp())` math to keep the scaling within bounds. With that we can make the list always use the full height of the overlay container which i have positioned to keep the recommended 5rem (=50px) space above and below.
- I moved the arrows above the list. I am not sure if you actually wanted that
- I added a `mask-image` to fade out the ambassador list
- I kept the previous scaling method for the card (using scale transforms) but added some more steps

Resolves #24

### DIFF
--- a/src/pages/overlay/components/activationButtons/activationButtons.module.scss
+++ b/src/pages/overlay/components/activationButtons/activationButtons.module.scss
@@ -1,19 +1,32 @@
 @import '../../../../variables.scss';
 
 .activationButtons {
-  margin-top: 40px;
+  margin-top: calc(6 * $overlay-base-size); // move slightly lower relative to available height
 
   button {
     all: unset;
+    display: flex;
+    justify-content: center;
+
+    // scale based on available height
+    width: calc(6.5 * $overlay-base-size);
+    height: calc(6.5 * $overlay-base-size);
+
     border-radius: 5px;
     background-color: $primary-color;
     box-shadow: 0 2px 4px $accent-color;
 
-    padding: 5px;
-
     img {
-      width: 55px;
-      height: 55px;
+      width: 80%;
+      height: auto;
+      aspect-ratio: 1;
+      object-fit: contain;
+
+      // the logo is not visually centered in the image
+      // so we tweak the position a bit to look centered
+      position: relative;
+      left: -2%;
+      top: -1%;
     }
 
     &:hover {
--- a/src/pages/overlay/index.scss
+++ b/src/pages/overlay/index.scss
@@ -5,6 +5,19 @@
   padding: 0;
 }
 
+// change default box-sizing mode to border-box
+// meaning total width/height includes padding and border
+*,
+::before,
+::after {
+  box-sizing: border-box;
+}
+
 body {
   font-family: $font-family;
+
+  // always keep full width/height and never show scrollbars
+  width: 100vw;
+  height: 100vh;
+  overflow: hidden;
 }
--- a/src/pages/panel/components/ambassadorPanel/AmbassadorPanel.tsx
+++ b/src/pages/panel/components/ambassadorPanel/AmbassadorPanel.tsx
@@ -46,6 +46,7 @@ export default function AmbassadorPanel() {
               src: ambassador.img.src,
               altText: ambassador.img.altText
             }}
+            ClassName={styles.item}
 
             getCard={()=>handleGetCard(ambassador.name)}
           />
--- a/src/utils/compositions/ambassador/ambassador.module.scss
+++ b/src/utils/compositions/ambassador/ambassador.module.scss
@@ -4,8 +4,7 @@
   display: flex;
   flex-direction: column;
   align-items: center;
-  justify-content: center;
-  padding: 20px 15px;
+  padding: 1.1em 0.9em;
   background-color: $primary-color;
   border-radius: 10px;
   text-align: center;
--- a/src/index.scss
+++ b/src/index.scss
@@ -1,4 +1,4 @@
-//Global styles
+// Global styles
 @import './variables.scss';
 
 * {
@@ -6,9 +6,18 @@
   padding: 0;
 }
 
+// change default box-sizing mode to border-box
+// meaning total width/height includes padding and border
+*,
+::before,
+::after {
+  box-sizing: border-box;
+}
+
 body {
   font-family: $font-family;
   color: $primary-text;
+
   display: flex;
   justify-content: center;
   overflow-x: hidden; // no horizontal scrollbar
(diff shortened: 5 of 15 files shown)


src/pages/overlay/index.scss
```diff
@@ -1,10 +1,40 @@
...
+    //but clamped between 6px and 10px to prevent tiny/giant UI
+    //the values are chosen so that in a normal desktop resolution
+    //1 base unit === 10px to make the math easier when using it
+    //to scale elements.
+    //
+    //usage example:
+    //  width: calc(5 * var(--alveus-base-size));
+    //  => which would be 50px in normal views and
+    //       would scale down to 30px
+    --alveus-base-size: clamp(6px, 1.46vh, 10px);
```
MattIPv4, line 14: Curious why this is being defined as a CSS var, rather than a SCSS var? Maybe personal pref, but with variables already defined in SCSS, I'd probably lean on a consistent approach to them?

pjeweb: Good catch. I tried to use media queries first to make the css variable dynamic (in steps), but moved to the clamp solution without media queries. I will update the PR later.

abdullahmorrison (changes requested):
The top of the overlay gets blocked by the Twitch header. It needs to be moved down a bit.

image

abdullahmorrison:
I like the arrow design 👍🏾

pjeweb:
> The top of the overlay gets blocked by the Twitch header. It needs to be moved down a bit.

Yeah I saw that when testing as well. The problem is that the title can be even bigger and will wrap to multiple (many) lines if the window is very small. And this only shows up in theatre mode or if embedded, so I just ignored that. But if you would prefer, I can move it down a bit. That is very easy to change with these changes.

Screenshot 2023-01-23 at 15 05 59

abdullahmorrison:
ye, on the extreme case it's fine, I just want it below the title if we assume there is not wrapping

src/pages/overlay/components/activationButtons/activationButtons.module.scss
```diff
@@ -1,16 +1,21 @@
 @import '../../../../variables.scss';
 
 .activationButtons{
-    margin-top: 40px;
+    margin-top: calc(6 * $overlay-base-size); //move slightly lower relative to available height
```
MattIPv4, line 4: Can this adjusted margin be copied over to overlaysettings to make sure they stay aligned?

pjeweb: yes sure. it is aligned at the top now, but otherwise it does not scale. i guess thats fine? the toggle button should not get any smaller, right?

MattIPv4: Thanks! That seems fine to me, I can't imagine scaling it will be useful (maybe at a certain point it should disappear?) If more settings are added down the road maybe it collapses into another menu like the list, but that's a future problem.

pjeweb:
PR is rebased and should leave enough space for the header, settings are aligned at the top.

Screenshot 2023-01-23 at 21 33 16

pjeweb:
Should I rebase this PR again or should I wait until you are sure you want to merge it?

abdullahmorrison:
I'd say wait for me to review it

src/pages/overlay/components/activationButtons/activationButtons.module.scss
```diff
@@ -1,16 +1,21 @@
...
     button{
         all: unset;
+        display: flex;
+        width: calc(5 * $overlay-base-size); //scale based on available height
```
abdullahmorrison, line 8: can you make the size of the button bigger. it feels a bit too small.

abdullahmorrison (changes requested):
The card information is to compacted. If we assume that the info for the ambassador story and conservation mission is full, there needs to be around 5px of margin between each row of the card. Also 5px at the bottom of the card.

image

image

src/utils/compositions/ambassador/ambassador.module.scss
```diff
@@ -6,7 +6,7 @@
     align-items: center;
     justify-content: center;
 
-    padding: 20px 15px;
+    padding: 7.7%;
```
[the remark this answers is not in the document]
pjeweb: the ambassador module does not get scaled using scale() anymore. to make the padding scale it must be relative or use the base size.  because the ambassador composition is being used for both overlay and panel, I do not use the base size but a relative size.

src/pages/panel/components/ambassadorPanel/AmbassadorPanel.tsx
```diff
@@ -41,17 +41,19 @@ export default function AmbassadorPanel() {
...
-            name={ambassador.name}
-            species={ambassador.species}
-            img={{
-              src: ambassador.img.src,
-              altText: ambassador.img.altText
-            }}
-
-            getCard={()=>handleGetCard(ambassador.name)}
-          />
+          <div className={styles.item}>
```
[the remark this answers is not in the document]
pjeweb: yeah the `ClassName` prop was added since the PR, can use that!

src/pages/overlay/components/activationButtons/activationButtons.module.scss
```diff
@@ -1,16 +1,21 @@
 @import '../../../../variables.scss';
 
 .activationButtons{
-    margin-top: 40px;
+    margin-top: calc(6 * $overlay-base-size); //move slightly lower relative to available height
     button{
```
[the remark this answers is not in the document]
pjeweb: its original size was not square. it was 65 x 71. which size would you want?

src/pages/overlay/components/activationButtons/activationButtons.module.scss
pjeweb: see https://github.com/abdullahmorrison/AlveusTwitchExtension/pull/26#discussion_r1088469243

src/pages/overlay/components/activationButtons/activationButtons.module.scss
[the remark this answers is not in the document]
pjeweb: It is 65x65 for now.

pjeweb:
Should all be changed as requested. I would hope we can merge it and work on it in separate PR/commits if necessary because the rebasing is very time consuming and error prone, as this PR touches so many files.

abdullahmorrison (approved):

PR #30, opened 2023-01-22
WIP map view concept

pjeweb:
Prototype of map view

Do not merge yet! There is a lot to fix still, but I would love some feedback!

Been working on a small map (like a zoo map) component that could be integrated into the extension.

Demo video: https://imgur.com/a/zZyXSKC


MattIPv4:
:eyes: I maintain a map of Alveus that was originally public, but at the request of Maya it has been made private for security.

pjeweb:
@MattIPv4 You think maya would have the same concerns with this map? Then I'll remove it as well

MattIPv4:
I don't want to speak on behalf of their team, but my map was less geographically accurate than this (to reduce the risk of leaking location etc., with the same idea of giving folks a visual representation of where all the ambassadors are on the site), and after a bit of back and forth it was decided to make it private (it's now used to keep track of all the pipe runs around the property for water/electric/network)

pjeweb:
To be clear my map is not at all geographically accurate as I drew it somewhere in the ocean. I just used a map library (and geojson) because that's what I am used to working with. Guess I'll remove it then, to be save.

MattIPv4:
Oh yah, I didn't mean geographically accurate in terms of location, but in terms of property layout etc.

Fwiw, it might be worth reaching out to Maya and the team to see if the view on having a map has changed, as I still think it'd be a really cool way to present things.

abdullahmorrison:
I think this is still a valid option. I can message the team to see their option on it. This needs to also be worked on for the panel as well if we commit to it.

pjeweb:
> I think this is still a valid option. I can message the team to see their option on it.

I mean us three seem to like the map idea, so if you would ask that would be great.

I am not that good of a designer or illustrator so I wanted to maybe ask the art chatters if they could help with some nice illustrations for the ambassadors and buildings (maybe even some trees or something). We could also make the map way more abstract/inaccurate and leave out more of the buildings/ways. But only if we get the blessing from the team :)

> This needs to also be worked on for the panel as well if we commit to it.

Should be doable in the panel as well. You can pretty much zoom in and out as much as you need. Might be cluttered in the most zoomed out view, but you could for example leave out some markers or cluster on the high zoom levels.

PR #39, opened 2023-01-23
Only connect to chat for the current channel

pjeweb:
This also moves the hardcoded test channel name to the `.env` file and move the `tmi.Client` initialization into the effect to prevent it from being recreated on every render.

Resolve #33


pjeweb:
Pretty happy with the PR now, if anyone has time to take a look at it. Managed to get jest tests running for the twitch api call and chat commands. I split up the environment variables so you can test with a test channel locally but build a production build without changing the vars. This adds a manual set up step. Maybe you could add a postinstall npm script for copying the `.env*.sample` files.

abdullahmorrison:
Wait for me to review this before adding to it. Then, if you want to add to it, create a comment first.

pjeweb:
> Wait for me to review this before adding to it. Then, if you want to add to it, create a comment first.

just removing the tests right now, did not realise you were reviewing right now, sorry

abdullahmorrison:
Also, make sure to name your branch `yourusername/your-feature` in the future

src/utils/twitch-api.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export async function fetchCurrentChannelInfo(channelId: string, auth: Twitch.ext.Authorized) {
+  const params = new URLSearchParams()
+  params.set('broadcaster_id', channelId)
+
+  const response = await fetch(`https://api.twitch.tv/helix/channels?${params}`, {
+    headers: {
+      'client-id': auth.clientId,
+      'Authorization': `Extension ${auth.helixToken}`,
```
MattIPv4, line 24: How does this behave if the user isn't signed in? Do you get a token that works anonymously?

src/utils/chatCommand.ts
```diff
@@ -35,41 +55,65 @@ export default function useChatCommand() {
...
+        setChannelInfo(newChannelInfo)
+      }
+    }
+
+    if ('Twitch' in window) {
+      window.Twitch.ext.onAuthorized(handleAuthorized)
+    }
+
+    return () => {
+      handleAuthorized = (_auth) => undefined
```
MattIPv4, line 93: 🤔 What is this doing? Reassigning here won't do anything, as a reference to the original method has already been passed to `onAuthorized`

abdullahmorrison:
FYI, I will add this functionality after the next version that I release of the extension so I won't be looking at this until then.

src/utils/chatCommand.ts
pjeweb: 1. good catch, somehow missed that i use the fn directly instead of `window.Twitch.ext.onAuthorized((auth) => handleAuthorized(auth))`
2. and I am not sure if this would even ever happen. It was supposed to "clean" up the async effect to prevent multiple calls from resolving out of order, but I guess this should never happen with this API. Using the effect with no deps (`[]`) also should pretty much guarantee it wont be called again. So probably fine to drop the cleanup fn.

MattIPv4: Yeah, the only reason really would be to clean up if this were to ever be unmounted -- I'd be inclined to say switching to `window.Twitch.ext.onAuthorized((auth) => handleAuthorized(auth))` is probably the better thing to do

src/utils/twitch-api.ts
pjeweb: This would most likely not work without a user sessopn. The callback does guarantee a valid `helixToken`, but would probably never be called without a twitch session. I could not test that out correctly using the twitch local extension preview. I will ask the twitch dev discord about it. That would also mean, we do not have the channel id. So it would not break break, but chat commands would not work for anons.

So we would need a fallback if no session is available. Not sure if it would be best to fall back to maya + alveussanctuary in that case or only alveussanctuary? Or we do use the streamer config view to let the alveus team configure it for each channel.

abdullahmorrison (changes requested):
Just need a default fallback for anon users. Then just rebase and I think it's good

pjeweb:
Default fallback is implemented. As recommended by the twitch dev discord I also moved the `onAuthorized` outside of the react component tree so it gets called immediately. The callback might only be called very early when React might not be ready rendering the components.

Instead of coding the fallback channels directly I also made them a environment variable (`REACT_APP_DEFAULT_CHANNEL_NAMES`).

pjeweb:
Oh and I could not figure out how to test the feature for anonymous users without submitting the extension for review. It does however work if you comment out the auth callback registration.

src/index.tsx
```diff
@@ -1,9 +1,12 @@
 import React from 'react';
 import ReactDOM from 'react-dom/client';
 import './index.css';
+import {registerTwitchAuthCallback} from './utils/channels'
 import App from './pages/panel/components/App';
 import reportWebVitals from './reportWebVitals';
 
+registerTwitchAuthCallback()
```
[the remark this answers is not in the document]
pjeweb: According to what I've read and been told in the twitch dev discord the onAuthorized callback we listen to might be called _before_ react components render. So an effect might be too late. That also depends on caching and other side effects so it's hard to be sure. That's the reason I made it so complicated and moved the registration outside of App.tsx.

abdullahmorrison (approved):

PR #40, opened 2023-01-24
Refactor ambassador commands to use fixed ids

pjeweb:
I do not known how important supporting the `!Jalapeño` command is, if not so much, i would propose this change: 
Instead of splitting and normalizing the name this defines the IDs in the `ambassadors.json` for each ambassador which makes it easier to work with the data. Is necessary you could also add some optional aliases here, if we need to support multiple spellings of one ambassador.

NOTE: this removes the option to use both names with and without diacritics in the commands (which is currently only !jalapeno).

### DIFF
--- a/src/assets/ambassadors.json
+++ b/src/assets/ambassadors.json
@@ -1,5 +1,6 @@
 [
     {
+        "id": "appa",
         "name": "Appa",
         "species": "Common Marmoset",
         "img": {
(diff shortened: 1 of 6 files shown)


abdullahmorrison (commented):
I agree that my solution is super overkill. But this is essentially hardcoding commands which I don't like the idea of. I want to have room to change what the commands are and possibly add different ways to write the same command without editing the json of every ambassador.

PR #47, opened 2023-01-30
Fix basic render tests

pjeweb:
This adds a test for the overlay app and makes them both pass basic rendering.


abdullahmorrison:
I don't want tests right now. I don't want to have to think about it and it won't make sense to have tests only in some places and not others.

abdullahmorrison:
make sure to create issues or discussion posts before contributing so you don't waste your time.

Issue #57, opened 2023-02-10
Make activation buttons stay highlighted when clicked


abdullahmorrison:
the activations buttons are only highlighted on hover. This should also be the case when they are clicked

MushkyQT:
Done, gonna open a PR

MushkyQT:
@abdullahmorrison Getting permission denied, could you please provide me with access rights to this repository?

MattIPv4:
You need to fork the repository and push to a branch there, then make a PR from your fork :) You don't need write access to this repo itself.

MushkyQT:
Ah got it, thanks @MattIPv4

PR #71, opened 2023-04-01
Make activation buttons stay highlighted when clicked

MushkyQT:
[Issue 57](https://github.com/abdullahmorrison/AlveusTwitchExtension/issues/57)

### DIFF
--- a/src/pages/overlay/components/activationButtons/ActivationButtons.tsx
+++ b/src/pages/overlay/components/activationButtons/ActivationButtons.tsx
@@ -16,10 +16,10 @@ interface ActivationButtonProps {
 export default function ActivationButtons(props: ActivationButtonProps) {
   return (
     <div className={styles.activationButtons}>
-        <button onClick={props.toggleShowAlveusIntro} className={`${props.isAlveusIntroActive && styles.highlighted}`}>
+        <button onClick={props.toggleShowAlveusIntro} className={`${props.isAlveusIntroActive ? styles.highlighted : ''}`}>
             <img src={Welcome} alt="Welcome" />
         </button>
-        <button onClick={props.toggleShowAmbassadorList} className={`${props.isAmbassadorListActive && styles.highlighted}`}>
+        <button onClick={props.toggleShowAmbassadorList} className={`${props.isAmbassadorListActive ? styles.highlighted : ''}`}>
             <img src={Ambassadors} alt="Ambassadors" />
         </button>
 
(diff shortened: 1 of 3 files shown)


src/pages/overlay/components/activationButtons/ActivationButtons.tsx
```diff
@@ -9,15 +9,17 @@ import Ambassadors from '../../../../assets/activationButtons/ambassadors.jpg';
...
   toggleShowAmbassadorList: () => void;
+  isAlveusIntroActive: boolean;
+  isAmbassadorListActive: boolean;
 }
 
 export default function ActivationButtons(props: ActivationButtonProps) {
   return (
     <div className={styles.activationButtons}>
-        <button onClick={props.toggleShowAlveusIntro}>
+        <button onClick={props.toggleShowAlveusIntro} className={`${props.isAlveusIntroActive && styles.highlighted}`}>
```
MattIPv4, line 19: Won't this result in `class="false"` when not active?

MushkyQT: Yes you're correct, fixed.
Thanks

MushkyQT: Do you know of any way to not define a className if there prop resolves to false? Currently, it will give the element `class=""`. I thought to render a different instance of the button without `className ` in a ternary just to avoid the empty class, but that feels a bit verbose

MattIPv4: You can set a prop to `undefined` if you don't want to render it normally

abdullahmorrison (changes requested):

src/pages/overlay/components/activationButtons/ActivationButtons.tsx
```diff
@@ -9,15 +9,17 @@ import Ambassadors from '../../../../assets/activationButtons/ambassadors.jpg';
...
   toggleShowAmbassadorList: () => void;
+  isAlveusIntroActive: boolean;
+  isAmbassadorListActive: boolean;
 }
 
 export default function ActivationButtons(props: ActivationButtonProps) {
   return (
     <div className={styles.activationButtons}>
-        <button onClick={props.toggleShowAlveusIntro}>
+        <button onClick={props.toggleShowAlveusIntro} className={`${props.isAlveusIntroActive ? styles.highlighted : ''}`}>
```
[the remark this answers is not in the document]
MushkyQT: Sweet thanks 👍

src/pages/overlay/components/activationButtons/ActivationButtons.tsx
MushkyQT: Got it 👍

abdullahmorrison (approved):

PR #79, opened 2023-05-11
Assorted ambassador card updates

### LINKED ISSUE #85: Add Ambassador Page link to Ambassador Card
adding a simple link to the Ambassador Card so one can navigate easier to the Ambassadors Page on the website

MattIPv4:
Resolves #85

Note that the moderator command hint will only work if the extension has the `Request Identity Link` capability enabled (though it will always work for the broadcaster).


MattIPv4:
@pjeweb would appreciate your eyes here as I refactored how the Twitch auth logic works -- `onAuthorized` seems to immediately call the given callback if its used after the initial auth has occurred, so I cleaned it up into a standard hook rather than requiring a top-level callback to register.

MattIPv4:
(@pjeweb would appreciate your review here as well as Abdullah, given that I've changed up how the auth stuff with Twitch works a bit which I believe you wrote)

src/utils/global/ambassadorCard/ambassadorCard.module.scss
```diff
@@ -1,16 +1,18 @@
...
-  width: 295px;
-  height: 460px;
-  max-height: 100vh;
+  width: 18.5rem;
+  max-width: 100%;
+  min-height: min(28.5rem, 100%);
+  max-height: 100%;
+
   justify-content: start;
   align-items: baseline;
```
pjeweb, line 10: I think this should be `flex-start`. Right now the content get offset to the right a bit.

src/utils/global/ambassadorCard/ambassadorCard.module.scss
```diff
@@ -1,16 +1,18 @@
 @import '../../../variables.scss';
 
 .ambassadorCard {
-  width: 295px;
-  height: 460px;
-  max-height: 100vh;
+  width: 18.5rem;
+  max-width: 100%;
+  min-height: min(28.5rem, 100%);
+  max-height: 100%;
```
pjeweb, line 7: `max-height` percents are based on the scaled height so it ends up less high than it needs to in small viewports. I came up with this workaround.  There probably is a better way though:

```scss
  @media screen and (max-height: 555px) {
    max-height: calc(100% / 0.9);
  }
  @media screen and (max-height: 500px) {
    max-height: calc(100% / 0.8);
  }
  @media screen and (max-height: 450px) {
    max-height: calc(100% / 0.7);
  }
  @media screen and (max-height: 410px) {
    max-height: calc(100% / 0.6);
  }
```

pjeweb (changes requested):
Looks good and seems to work well. The refactoring seems solid. I like the mod hints!

Can we make the scrollbar look nicer? (see https://css-tricks.com/the-current-state-of-styling-scrollbars-in-css/)

```scss
.scrollable {
   …
    margin-right: 1px;

    // For Chrome & Safari
    &::-webkit-scrollbar {
      width: 6px;
      height: 6px;
    }
    &::-webkit-scrollbar-thumb {
      background: $primary-color;
      border-radius: 4px;
    }
    &::-webkit-scrollbar-track {
      background: $accent-color;
    }

    // Standard version (Firefox only for now)
    scrollbar-color: $primary-color $accent-color;
  }
}
```

src/utils/global/ambassadorCard/ambassadorCard.module.scss
MattIPv4: I went with `align-items: start` -- I was developing in Firefox where it seems this wasn't an issue, but yeah, in Chrome it was misaligned with the scrollbar -- very odd.

src/utils/global/ambassadorCard/ambassadorCard.module.scss
MattIPv4: As discussed in Discord, not going to fix this in this PR, but will open an issue/PR to revisit scaling holistically, with the ideal being to remove the use of scale and instead change the font-size at breakpoints, with the whole extension using rems so everything scales accordingly.

pjeweb (approved):

pjeweb (approved):

PR #91, opened 2023-05-28
Fix overlay responsiveness

### LINKED ISSUE #88: Revisit extension scaling, use font-size + rems, not scale
As discussed in Discord, not going to fix this in this PR, but will open an issue/PR to revisit scaling holistically, with the ideal being to remove the use of scale and instead change the font-size at breakpoints, with the whole extension using rems so everything scales accordingly.

_Originally posted by @MattIPv4 in https://github.com/alveusgg/extension/pull/79#discussion_r1206086968_

MattIPv4:
Resolves #88


MattIPv4:
Lmk how this behaves for you, and lmk if you have any concerns with the panel scaling (it seemed fine to me, but maybe a reduced version of stepping the font-size down there might be useful too?)

pjeweb (approved):
does work well for me :)

pjeweb:
to be a bit more specific: the main problem with the card being cut off is solved, but the list could fill out the available space better in some cases. i don't think this i a big problem though. we might want to implement some sort of filtering/sorting/grouping for the panel view at some point?! and a button to show the welcome info...

MattIPv4:
> but the list could fill out the available space better in some cases.

How so? We're not meant to fill the reserved regions at the top/bottom, and it is set to be using 100% of the height that it can?

> we might want to implement some sort of filtering/sorting/grouping for the panel view at some point?! and a button to show the welcome info...

Yeah, I think that'd be interesting to play around with -- maybe grouping by enclosure like we do on the site?

pjeweb:
> > but the list could fill out the available space better in some cases.
> 
> 
> 
> How so? We're not meant to fill the reserved regions at the top/bottom, and it is set to be using 100% of the height that it can?


I was specifically referring to the panel view, not the overlay. I think we can use the full size there.


> > we might want to implement some sort of filtering/sorting/grouping for the panel view at some point?! and a button to show the welcome info...
> 
> 
> 
> Yeah, I think that'd be interesting to play around with -- maybe grouping by enclosure like we do on the site?

Yeah something like that would be great. (Also of course I still want to sync with the OBS scene what ambassadors are "active" ;) )

Issue #25, opened 2023-01-21
Add tool tips (CSS only)


abdullahmorrison:
Tooltips will pop up on hover.
- [ ] One next to activation buttons 

<img src="https://user-images.githubusercontent.com/49528805/213881529-2cae4fbe-aafa-4ec2-8d40-2fe383598468.png" width=300/>

- [ ] One inside an ambassador card, next to the IUCN status that explains what an IUCN status is 

<img src="https://user-images.githubusercontent.com/49528805/213882521-0d64dcf5-85ae-4a1e-99a0-fc10cd189160.png" width = 500/>

GersonMJL:
Yo, I'll try to implement this

ThatGravyBoat:
As a temporary change for the buttons at least the [`title` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) would be excellent while tooltips are being developed.

MushkyQT:
@GersonMJL Hey, are you still planning on implementing the tooltips? If not just lmk, I'll have a go

GersonMJL:
@MushkyQT Yeah you can have it, I don't have the time rn :/

MushkyQT:
@GersonMJL Sounds good!

MattIPv4:
👋 @MushkyQT Any progress on this, or should I unassign it so someone else can take a look?

jibranabsarulislam:
^Please let me know the status, I'd be down to take a look!

MattIPv4:
Sure, I'll re-assign this to you :)

MattIPv4:
#101

PR #101, opened 2023-06-04
Implement tooltips for overlay buttons + conservation status

jibranabsarulislam:
#25
Custom Tooltip component to match the design spec

i'm definitely open to:

- [x] ~~1) fix the spaghetti Tooltip component style (i call it "_dynamic_")~~ done
- [ ] 2) put together the tooltip shape for overlay buttons (different from current/default)
- [x] ~~3) `zIndex `changes? I couldn't get the hover text to show up on top of the ambassador components even when I set each style to 3 or above~~ done

https://github.com/alveusgg/extension/assets/70596906/c7999fc1-6d32-435c-ab11-16f62bdaa8c8


src/pages/overlay/components/buttons/Buttons.tsx
```diff
@@ -1,10 +1,11 @@
-import { useMemo } from 'react'
+import { useMemo, useState } from "react";
```
MattIPv4, line 1:
```suggestion
import { useMemo, useState } from 'react'
```
Nitpick, imports here appear to use single quotes with no semi, let's stick to that

src/pages/overlay/components/buttons/Buttons.tsx
```diff
@@ -1,10 +1,11 @@
...
 
 import { classes } from '../../../../utils/classes'
 
 import styles from './buttons.module.scss'
+import Tooltip from "../tooltip/Tooltip";
```
MattIPv4, line 6:
```suggestion
import { classes } from '../../../../utils/classes'
import Tooltip from '../tooltip/Tooltip'

import styles from './buttons.module.scss'
```
Nitpick, import sorting + quotes/semi consistency

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -34,20 +34,23 @@ const overlayOptions = [
...
     icon: WelcomeIcon,
+    hoverText: 'Welcome',
     component: WelcomeOverlay,
   },
   {
     key: 'ambassadors',
     title: 'Ambassadors',
     type: 'primary',
     icon: AmbassadorsIcon,
+    hoverText: 'List of Ambassadors',
```
MattIPv4, line 45:
```suggestion
    hoverText: 'Explore our Ambassadors',
```

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -34,20 +34,23 @@ const overlayOptions = [
     title: 'Welcome',
     type: 'primary',
     icon: WelcomeIcon,
+    hoverText: 'Welcome',
```
MattIPv4, line 37:
```suggestion
    hoverText: 'Welcome to Alveus',
```

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -34,20 +34,23 @@ const overlayOptions = [
...
     icon: AmbassadorsIcon,
+    hoverText: 'List of Ambassadors',
     component: AmbassadorsOverlay,
   },
   {
     key: 'settings',
     title: 'Settings',
     type: 'secondary',
     icon: SettingsIcon,
+    hoverText: 'Settings',
```
MattIPv4, line 53:
```suggestion
    hoverText: 'Extension Settings',
```

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -1,3 +1,5 @@
+import {useState} from 'react'
```
MattIPv4, line 1:
```suggestion
import { useState } from 'react'
```

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -7,6 +9,8 @@ import { classes } from '../../classes'
 import styles from './ambassadorCard.module.scss'
 import moderatorBadge from '../../../assets/mod.png'
 
+import Tooltip from '../../../pages/overlay/components/tooltip/Tooltip'
```
MattIPv4, line 12: I think tooltip should probably be in utils/global, given that we're using it in a global component here

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -19,6 +23,29 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
...
+    {/* svg sourced from https://icons.getbootstrap.com/icons/info-circle-fill/. The Tailwind-recommended https://heroicons.com/ wasn't friendly for resizing */}
+    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="gray" viewBox="0 0 16 16">
+      <path d="M8 16A8 8 0 1 0 8 0a8 8 0 0 0 0 16zm.93-9.412-1 4.705c-.07.34.029.533.304.533.194 0 .487-.07.686-.246l-.088.416c-.287.346-.92.598-1.465.598-.703 0-1.002-.422-.808-1.319l.738-3.468c.064-.293.006-.399-.287-.47l-.451-.081.082-.381 2.29-.287zM8 5.5a1 1 0 1 1 0-2 1 1 0 0 1 0 2z"/>
+    </svg>
+    <Tooltip
+      text={iucnText}
+      isShown={iucnTooltipShown}
+      offsetLeft={11}
+      offsetTop={16.75}
+      width={18}
```
MattIPv4, line 44: These values seem very, very specific -- is there not a better way to position this that's just based on the parent etc.?

MattIPv4 (changes requested):
Hm, what're your thoughts on Tooltip taking children, so that you just wrap an item in the tooltip component. That way, the tooltip component can manage detecting hovering etc. without needing to pollute every implementation with the same event listeners?

Having it wrap the children would also allow you to set a `title` prop on the wrapper div, making this more accessibiliy-friendly :)

I would also agree that it'd be nice to have this use scss module styles like everything else, I'm generally not a fan of mixing different ways of achieving the same thing in a project, it causes confusion down the road etc.

MattIPv4:
> 3\. `zIndex `changes? I couldn't get the hover text to show up on top of the ambassador components even when I set each style to 3 or above
>         https://github.com/alveusgg/extension/assets/70596906/c7999fc1-6d32-435c-ab11-16f62bdaa8c8

That's odd -- I believe the overlay is structured such that the active overlay section/card is always below the buttons. Perhaps you need a z-index: 0 on the parent of the buttons? If not, I can pull down and play around with getting the stacking correct.

src/utils/global/ambassadorCard/AmbassadorCard.tsx
jibranabsarulislam: i agree I really tweaked it by the decimal there 😅

i'll think of something. The reason I input those values particularly is since we'd be sharing the Tooltip functionality with the overlay buttons, it's difficult to set position based on percentages because the overlay buttons's 100% evaluates to less than the ambassador's 100%.

thanks a ton for your feedback! I'll get to the changes and implementing Tooltip as a wrapper

src/pages/overlay/components/overlay/ambassadors/ambassadors.module.scss
```diff
@@ -6,7 +6,7 @@
   left: 0;
   height: 100%;
   display: flex;
-
+  z-index: 0; // lets tooltips appear above the list
```
MattIPv4, line 9:
```suggestion
  z-index: 0; // lets tooltips appear above the list

```
nitpick, linebreak between blocks

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -7,6 +9,8 @@ import { classes } from '../../classes'
 import styles from './ambassadorCard.module.scss'
 import moderatorBadge from '../../../assets/mod.png'
 
+import Tooltip from '../tooltip/Tooltip'
```
MattIPv4, line 12: Nitpick, this should be sorted above the styles/assets imports, with the chatCommand import etc.

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,39 @@
...
+    opacity: props.isShown ? 1 : 0,
+    width: `${props.width}rem` || "auto",
+  };
+
+  const tooltipClass = props.isOption
+    ? styles.optionStyles
+    : styles.defaultStyles;
+
+  return (
+    <>
```
MattIPv4, line 26: I would have this wrap the tooltip and its children in a div -- that way, this component can handle the hover detection, so that it becomes an isolated component and you aren't polluting everywhere that uses it with hover detection logic

src/utils/global/tooltip/tooltip.module.scss
```diff
@@ -0,0 +1,35 @@
+$tooltip-background-color: #18181B;
+
+.sharedStyles {
+  background-color: $tooltip-background-color;
+  border-radius: 0.5rem;
+  color: white;
+  padding: 0.5rem;
+  position: absolute;
+  text-align: left;
+  transition: opacity 0.15s ease-in-out;
```
MattIPv4, line 10: Could we use the standard `$transition` here?

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,39 @@
...
+  isOption?: boolean;
+  offsetLeft?: number;
+  offsetTop?: number;
+  width?: number;
+}
+
+const Tooltip: React.FC<TooltipProps> = (props) => {
+  const tooltipStyle: React.CSSProperties = {
+    opacity: props.isShown ? 1 : 0,
+    width: `${props.width}rem` || "auto",
```
MattIPv4, line 18: This doesn't seem quite right? `"undefinedrem"` is going to be truthy, so `|| "auto"` will never apply.

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+    opacity: show ? 1 : 0,
+    width: props.textContainerWidth ? props.textContainerWidth : "max-content",
+    left: props.leftOffset,
+  };
+
+  return (
+    <>
+      <div
+        className={styles.wrapper}
+        // title={props.title}
```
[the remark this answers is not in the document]
MattIPv4: Ah yeah, that makes sense -- yeah, aria-label would be great

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -19,6 +22,24 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
   const images = getAmbassadorImages(ambassadorKey)
   const mod = window?.Twitch?.ext?.viewer?.role === 'broadcaster' || window?.Twitch?.ext?.viewer?.role === 'moderator'
 
+
+  const iucnText = "An objective assessment system for classifying the status of plants, animals, and other organisms threatened with extinction."
+  const iucnInfoIcon = (
+    <Tooltip
+      text={iucnText}
+      title="IUCN"
+      leftOffset="60%"
```
jibranabsarulislam, line 31: this is still necessary because I tried to `position: relative` the wrapper and children and `position: absolute` the tooltip, which worked for the buttons

but, in the case of the ambassador card, this caused the card to contain the tooltip resulting in an overflow-x instead of using the `position: absolute` style which i thought should ignore its container and allow position based on its parent

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -19,6 +22,24 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
...
+      textContainerWidth="18rem"
+    >
+      <button>
+        {/* svg sourced from https://icons.getbootstrap.com/icons/info-circle-fill/. The Tailwind-recommended https://heroicons.com/ wasn't friendly for resizing */}
+        <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="gray" viewBox="0 0 16 16">
+          <path d="M8 16A8 8 0 1 0 8 0a8 8 0 0 0 0 16zm.93-9.412-1 4.705c-.07.34.029.533.304.533.194 0 .487-.07.686-.246l-.088.416c-.287.346-.92.598-1.465.598-.703 0-1.002-.422-.808-1.319l.738-3.468c.064-.293.006-.399-.287-.47l-.451-.081.082-.381 2.29-.287zM8 5.5a1 1 0 1 1 0-2 1 1 0 0 1 0 2z"/>
+        </svg>
+      </button>
+    </Tooltip>
+  );
```
MattIPv4, line 41: Is there a reason for having this as a const, rather than just having it directly in the JSX below?

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -1,12 +1,15 @@
+
```
MattIPv4, line 1:
```suggestion
```

src/utils/global/ambassadorCard/AmbassadorCard.tsx
jibranabsarulislam: right, I think your suggestion makes sense; i'll push the refactor shortly

my intuition was to declare it so-as to not clutter the return, but I see that since it's a component and not logic that it should reside within the return statement

src/utils/global/tooltip/tooltip.module.scss
```diff
@@ -0,0 +1,39 @@
...
+
+.children {
+  position: relative;
+  display: flex;
+  height: 100%;
+  width: 100%;
+  align-items: center;
+  justify-content: center;
+}
+.wrapper {
```
MattIPv4, line 33:
```suggestion

.wrapper {
```
Nitpick, linebreak between blocks

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -3,10 +3,12 @@ import { getAmbassadorImages, getIUCNStatus, type AmbassadorKey, type Ambassador
 import { normalizeAmbassadorName } from '../../chatCommand'
 import { camelToKebab } from '../../helpers'
 import { classes } from '../../classes'
+import Tooltip from '../tooltip/Tooltip'
 
 import styles from './ambassadorCard.module.scss'
 import moderatorBadge from '../../../assets/mod.png'
 
+
```
MattIPv4, line 11:
```suggestion
```

src/utils/global/tooltip/tooltip.module.scss
```diff
@@ -0,0 +1,39 @@
...
+  display: inline-flex;
+  height: 100%;
+  width: 100%;
+  align-items: center;
+  justify-content: center;
```
MattIPv4, line 38:
```suggestion
  position: relative;
```

I'm not sure any of this should be default styling really -- I think the only thing I'd expect here is a `position: relative` to ensure the tooltip is positioned correctly.

I would expect the Tooltip to accept a className prop though, which gets applied to the wrapper -- this would allow specific uses to pass styles as needed, such as width/height 100%.

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <>
+      <div
+        className={styles.wrapper}
+        aria-label={props.title}
+        onMouseEnter={() => setShow(true)}
+        onMouseLeave={() => setShow(false)}
+      >
+        <div className={styles.children}>{props.children}</div>
```
MattIPv4, line 30:
```suggestion
        {props.children}
```
The children shouldn't need to be wrapped in another div, that seems unnecessary and just adding extra DOM levels?

src/utils/global/tooltip/tooltip.module.scss
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.children {
+  position: relative;
+  display: flex;
+  height: 100%;
+  width: 100%;
+  align-items: center;
+  justify-content: center;
+}
```
MattIPv4, line 32:
```suggestion
```
See related comment about not needing to wrap the children

src/utils/global/tooltip/tooltip.module.scss
```diff
@@ -0,0 +1,39 @@
...
+.sharedStyles {
+  background-color: $tooltip-background-color;
+  border-radius: 0.5rem;
+  color: white;
+  font-size: 1rem;
+  left: calc(100% + .75rem);
+  padding: 0.5rem;
+  position: absolute;
+  transition: $transition;
+
```
MattIPv4, line 12:
```suggestion
  transition-property: opacity;
```
Can we specify which properties we're expecting to transition? I'm assuming just opacity?

src/utils/global/ambassadorCard/ambassadorCard.module.scss
```diff
@@ -70,6 +70,12 @@
     margin: 0 0.5rem 0.25rem;
   }
 
+  .col {
+    display: inline-flex;
+    flex-direction: column;
+    width: fit-content;
+  }
```
MattIPv4, line 77:
```suggestion
  .info {
    display: flex;
    align-items: center;
    gap: .5rem;
    justify-content: space-between;
    
    .tooltip {
      button,
      svg {
        display: block;
      }
    }
  }
```
Your current approach seemed to have some alignment issues, hopefully this cleans it up (see my other comment with the related JSX changes)

src/utils/global/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -73,7 +75,30 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
...
+          <div className={styles.col}>
+            <h3>Conservation Status</h3>
+          </div>
+          <div className={styles.col}>
+            <Tooltip
+              text="An objective assessment system for classifying the status of plants, animals, and other organisms threatened with extinction."
+              title="IUCN"
+              leftOffset="60%"
+              textContainerWidth="18rem"
+            >
```
MattIPv4, line 87:
```suggestion
          <div className={styles.info}>
            <h3>Conservation Status</h3>
            <Tooltip
              text="An objective assessment system for classifying the status of plants, animals, and other organisms threatened with extinction."
              title="IUCN"
              leftOffset="60%"
              textContainerWidth="18rem"
              className={styles.tooltip}
            >
```
See related comment with the SCSS changes, and this also relies on my other suggestion about support a className in tooltip that is passed to the wrapper

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const [show, setShow] = useState(false);
+
+  const tooltipStyle: React.CSSProperties = {
+    opacity: show ? 1 : 0,
+    width: props.textContainerWidth ? props.textContainerWidth : "max-content",
+    left: props.leftOffset,
+  };
+
+  return (
+    <>
```
MattIPv4, line 23: Fragment isn't needed here, as you're returning a single div element as the root

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <div
+        className={styles.wrapper}
```
MattIPv4, line 25: Per my other comments/suggestions, I think this should also accept an optional className prop that is passed in here -- we have a util that'll do that cleanly, `classes`

src/utils/global/tooltip/tooltip.module.scss
```diff
@@ -0,0 +1,39 @@
+@import '../../../variables.scss';
+
+.sharedStyles {
+  background-color: $tooltip-background-color;
+  border-radius: 0.5rem;
+  color: white;
+  font-size: 1rem;
+  left: calc(100% + .75rem);
+  padding: 0.5rem;
+  position: absolute;
```
MattIPv4, line 10:
```suggestion
  pointer-events: none;
  position: absolute;
```
I think you probably want to set `pointer-events: none` on the tooltip -- currently, if you hover where the tooltip will be, it will show it.

src/utils/global/tooltip/tooltip.module.scss
MattIPv4: I'm realising that using `position: relative` might cause some issues with the conservation status tooltip, as it then forces it to be positioned within the scrollable container -- off the top of top of my head I can't think of a great solution to this.

Currently, the position completely breaks when the card starts scrolling, but with `position: relative` it will be broken in a different way.

image

MattIPv4: I'm not sure if there's a good CSS-only solution to this, so might need a bit of JavaScript?

https://css-tricks.com/popping-hidden-overflow/ goes over a very similar problem and lands on JavaScript as the solution.

src/utils/global/tooltip/Tooltip.tsx
MattIPv4: > something with the passing the className as a prop, would the tooltip component have access to those styles through its module.scss? I'm not sure if it's setup using the .module.scss file extension.

Yeah it should work just fine

jibranabsarulislam:
@MattIPv4 this one hit me like a final boss 😵

using ref seemed to work. it was a result of a lot of searching, trial-and-error (and copilot), but the tooltip should show up to the right of any element as long as it has the appropriate className and children. I want to look more into how libraries like MUI are able to position their tooltips and handle the cases like we had here with enclosing containers

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+      >
+        {props.children}
+      </div>
+      <div
+        className={styles.tooltip}
+        ref={tooltipRef}
+        style={{
+          ...tooltipStyle,
+          top: containerCenter - (tooltipRect?.height ?? 0) / 2,
+          left: (containerRef.current?.getBoundingClientRect().right ?? 0) + 10,
```
MattIPv4, line 55:
```suggestion
          left: (containerRect.right ?? 0) + 10,
```
Don't need to re-compute this

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+    (containerRect?.top ?? 0) + (containerRect?.height ?? 0) / 2;
+
+  // set tooltip style based on incoming props
+  const tooltipStyle = {
+    opacity: show ? 1 : 0,
+    // restrict width of tooltip with optional prop
+    width: props.textContainerWidth ? props.textContainerWidth : "auto",
+    fontSize: props.fontSize ? props.fontSize : "1rem",
+  };
+
```
MattIPv4, line 35:
```suggestion

  // add event listeners to the children to show/hide tooltip
  const childrenWithProps = React.Children.map(props.children, (child) => {
    if (React.isValidElement(child)) {
      return React.cloneElement(child as React.ReactElement, {
        onMouseEnter: () => setShow(true),
        onMouseLeave: () => setShow(false),
      });
    }
    return child;
  });

```
To avoid having to wrap the children in two additional layers, we can add the event bindings directly to the children

src/utils/global/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+      ref={containerRef}
+      aria-label={props.ariaLabel}
+    >
+      <div
+        onMouseEnter={() => setShow(true)}
+        onMouseLeave={() => setShow(false)}
+        className={classes(props.className, styles.children)}
+      >
+        {props.children}
+      </div>
```
MattIPv4, line 48:
```suggestion
      {childrenWithProps}
```
See related suggestion

src/pages/overlay/components/buttons/Buttons.tsx
```diff
@@ -27,15 +28,30 @@ export default function Buttons<T extends ButtonsOptions = ButtonsOptions>(props
...
+      {optionsWithOnClick.map((option) => (
+        <Tooltip
           key={option.key}
-          onClick={option.onClick}
-          className={classes(option.active && styles.highlighted, option.type === "secondary" && styles.secondary)}
+          text={option.hoverText}
+          ariaLabel={option.title}
+          className={classes(
+            styles.tooltip,
+            option.active && styles.highlighted,
```
MattIPv4, line 38:
```suggestion
```
We don't need to double up the highlighting

src/pages/overlay/components/buttons/Buttons.tsx
```diff
@@ -27,15 +28,30 @@ export default function Buttons<T extends ButtonsOptions = ButtonsOptions>(props
...
-        <button
+      {optionsWithOnClick.map((option) => (
+        <Tooltip
           key={option.key}
-          onClick={option.onClick}
-          className={classes(option.active && styles.highlighted, option.type === "secondary" && styles.secondary)}
+          text={option.hoverText}
+          ariaLabel={option.title}
+          className={classes(
+            styles.tooltip,
```
MattIPv4, line 37:
```suggestion
            styles.btn,
```
No point having two classes doing the same thing

src/pages/overlay/components/buttons/Buttons.tsx
```diff
@@ -27,15 +28,30 @@ export default function Buttons<T extends ButtonsOptions = ButtonsOptions>(props
...
         >
-          <img src={option.icon} alt={option.title} />
-        </button>
+          <button
+            key={option.key}
+            onClick={option.onClick}
+            className={classes(
+              styles.btn,
+              option.active && styles.highlighted,
+              option.type === "secondary" && styles.secondary
```
MattIPv4, line 48:
```suggestion
              option.active && styles.highlighted,
```
We only need to worry about the highlighting here

src/pages/overlay/components/buttons/buttons.module.scss
```diff
@@ -7,27 +7,31 @@
   margin: 3rem 0 0;
   z-index: 1; // stay above the popups
 
-  button {
+  .btn,
+  .tooltip {
     display: flex;
     justify-content: center;
     align-items: center;
```
MattIPv4, line 14:
```suggestion
  .btn {
```
No point having two classes doing the same thing + moving the flex alignment to the button, as the image is inside that

src/pages/overlay/components/buttons/buttons.module.scss
```diff
@@ -7,27 +7,31 @@
...
-    background: $primary-color;
-    cursor: pointer;
-    box-shadow: 0 0.125rem 0.25rem $accent-color;
-    transition: $transition;
-    transition-property: outline, filter;
 
-    img {
-      height: 60%;
-    }
+    button {
```
MattIPv4, line 18:
```suggestion
    button {
      width: 100%;
      height: 100%;
      display: flex;
      justify-content: center;
      align-items: center;
```
The button is what contains the image, and so should have the flex alignment

src/pages/overlay/components/buttons/Buttons.tsx
```diff
@@ -27,15 +28,30 @@ export default function Buttons<T extends ButtonsOptions = ButtonsOptions>(props
...
+          className={classes(
+            styles.tooltip,
+            option.active && styles.highlighted,
+            option.type === "secondary" && styles.secondary
+          )}
         >
-          <img src={option.icon} alt={option.title} />
-        </button>
+          <button
+            key={option.key}
```
MattIPv4, line 43:
```suggestion
```
This doesn't need a key, it's not being mapped

src/pages/overlay/components/buttons/buttons.module.scss
jibranabsarulislam: ah, thanks a ton again Matt this makes 100% sense. I was getting some problems with the image not taking the correct size inside the tooltip, but the height and width 100 along with the flex was the solution

MattIPv4 (approved):

MattIPv4:
Thanks for all the work on this 💚

Issue #80, opened 2023-05-18
Hide cursor on idle


abdullahmorrison:
make the cursor disappear while on the overlay and idle so it doesn't block the video

SUSINDRAREDDY:
Hey I like your project, I like to contribute in this issue. 
Is it up for grabs?

abdullahmorrison:
Go ahead, I'll assign it to you

SUSINDRAREDDY:
Can you provide me more details to address this issue

![image](https://github.com/alveusgg/extension/assets/126981466/607e9f69-868f-47a0-9b6c-5741df742216)
 Is this issue for full screen?

MattIPv4:
👋 Hey @SUSINDRAREDDY, have you had a chance to get started on this? Anything we can do to help?

MattIPv4:
No movement, unassigning so someone else can take a look at tackling this if they wish.

jibranabsarulislam:
hey all, I've got this one working on my branch. with your permission I'll open a PR w more info

MattIPv4:
Sure 👍

MattIPv4:
#102

PR #102, opened 2023-06-06
Style cursor to none when overlay is idle

jibranabsarulislam:
#80 
some caveats to address:
- [ ] `cursor: none` applies only when window is focused
- [ ] overlay container is smaller than total size of screen, so the cursor shows around the edges in local dev (intended?)

tested in Chromium (brave, chrome), firefox, edge

https://github.com/alveusgg/extension/assets/70596906/ff248213-a040-4400-8159-d5dab27229f9

### DIFF
--- a/src/pages/overlay/App.tsx
+++ b/src/pages/overlay/App.tsx
@@ -90,7 +90,7 @@ export default function App() {
 
   // When the user interacts, have a 5s timeout before hiding the overlay
   const interacted = useCallback(() => {
-    wake(5000)
+    wake(1000)
   }, [wake])
 
   // When a user scrolls, treat it as an interaction (but handle Firefox being weird)
--- a/src/pages/overlay/App.tsx
+++ b/src/pages/overlay/App.tsx
@@ -90,7 +90,7 @@ export default function App() {
 
   // When the user interacts, have a 5s timeout before hiding the overlay
   const interacted = useCallback(() => {
-    wake(1000)
+    wake(5000)
   }, [wake])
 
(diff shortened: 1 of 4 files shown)


MattIPv4:
Instead of doing this on the `.hidden` class, which will have the gaps you observed due to us following Twitch's recommended dead zones, what about adding a `useEffect` in `App.tsx` that's driven by `sleeping`, and applies this style directly to the HTML element, so that it affects the entire overlay?

jibranabsarulislam:
https://github.com/alveusgg/extension/assets/70596906/d11eb28d-d723-49d9-b8bc-f8038aa971e4

src/pages/overlay/App.tsx
```diff
@@ -93,6 +93,12 @@ export default function App() {
     wake(5000)
   }, [wake])
 
+  useEffect(() => {
+    if (sleeping) document.documentElement.style.cursor = 'none'
+    else document.documentElement.style.cursor = 'default'
+  }, [sleeping])
+
+
```
MattIPv4, line 101:
```suggestion
  // Hide the cursor when sleeping
  useEffect(() => {
    if (sleeping) document.documentElement.style.cursor = 'none'
    else document.documentElement.style.removeProperty('cusor')
  }, [sleeping])

```
Few nitpicks here -- comment to describe what the effect is doing, using `removeProperty` to cleanup rather than setting it explicitly to `"default"`, and removing an extra linebreak

MattIPv4 (approved):

MattIPv4:
Thank you for your work on this, much appreciated! 💚

PR #104, opened 2023-06-09
Move core app logic to proper hooks/contexts

MattIPv4:
Should be no real functionality change, just cleaning up a lot of the app logic that was reliant on passing props everywhere into proper hooks


pjeweb (approved):
Clean

PR #105, opened 2023-06-09
Implement eslint + prettier w/ husky + lint-staged hook

### LINKED ISSUE #100: Implement prettier w/ lint-staged + husky
Implement prettier across the whole repo, default settings.

Using husky + lint-staged, ensure that a pre-commit hook exists for prettier, running on all files.

Remove `lint:editorconfig`, replace with `lint:prettier`. Add `lint:prettier:fix`.

Ensure eslint is happy running alongside prettier.

jibranabsarulislam:
Resolves #100 

runs eslint then prettier on project files. husky's in play on commits

to commit the config without the file changes in 842cd8a4955b7ea840d6d7456a3cfb306f22ba8c i used `git commit -m "..." --no-verify`

would the lint step need to be called in the workflow file for testing PRs? https://github.com/alveusgg/extension/blob/ad1c843db26fb46a06a2ba36f80c813a99c6f7ce/.github/workflows/test.yml


package.json
```diff
@@ -62,11 +61,27 @@
...
+    "prettier": "^2.8.8",
     "react-app-rewire-multiple-entry": "^2.2.2",
     "react-app-rewired": "^2.2.1",
     "typescript": "^4.9.5",
     "webpack-image-resize-loader": "^5.0.0"
+  },
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx,json}": [
+      "npm run lint:fix",
+      "npm run format"
```
MattIPv4, line 79: We want to call eslint/prettier directly here, otherwise you're going to be running them against all files every time, defeating the point of using lint-staged

package.json
```diff
@@ -27,10 +27,9 @@
     "start": "react-app-rewired start",
     "build": "react-app-rewired build",
     "test": "react-app-rewired test",
-    "lint": "npm run lint:editorconfig && npm run lint:eslint",
-    "lint:editorconfig": "editorconfig-checker",
-    "lint:eslint": "eslint --ext .js,.jsx,.ts,.tsx src",
-    "lint:eslint:fix": "npm run lint:eslint -- --fix",
+    "lint": "eslint src --ext .ts,.tsx",
+    "lint:fix": "eslint src --ext .js,.jsx,.ts,.tsx --fix",
+    "format": "prettier --write \"src/**/*.{js,jsx,ts,tsx,json}\"",
```
MattIPv4, line 32:
```suggestion
    "lint:eslint": "eslint src --ext .ts,.tsx",
    "lint:prettier": "prettier --check \"src/**/*.{js,jsx,ts,tsx,json}\"",
    "format:eslint": "eslint src --ext .js,.jsx,.ts,.tsx --fix",
    "format:prettier": "prettier --write \"src/**/*.{js,jsx,ts,tsx,json}\"",
```
We want our lint/format scripts to be able to run both (might be worth adding npm-run-all to expose lint/format to run both)

package.json
```diff
@@ -27,10 +27,9 @@
...
     "eject": "react-scripts eject"
   },
   "eslintConfig": {
```
MattIPv4, line 35: There's an existing eslint config in here that needs to be addressed

.prettierrc.json
MattIPv4, line 1: I don't think we need this file as we're not overriding anything?

package.json
```diff
@@ -62,11 +58,28 @@
     "@types/tmi.js": "^1.8.2",
     "@types/twitch-ext": "^1.24.4",
     "editorconfig-checker": "^5.0.1",
-    "eslint": "^8.33.0",
+    "eslint": "^8.42.0",
+    "eslint-config-prettier": "^8.8.0",
     "eslint-config-react-app": "^7.0.1",
```
MattIPv4, line 63: It looks like your config is no longer using this, so can we remove this dep if we don't need it?

package.json
```diff
@@ -27,18 +27,14 @@
     "start": "react-app-rewired start",
     "build": "react-app-rewired build",
     "test": "react-app-rewired test",
-    "lint": "npm run lint:editorconfig && npm run lint:eslint",
-    "lint:editorconfig": "editorconfig-checker",
-    "lint:eslint": "eslint --ext .js,.jsx,.ts,.tsx src",
-    "lint:eslint:fix": "npm run lint:eslint -- --fix",
+    "lint:eslint": "eslint src --ext .ts,.tsx",
+    "lint:prettier": "prettier --check \"src/**/*.{js,jsx,ts,tsx,json}\"",
+    "lint": "npm-run-all lint:*",
```
MattIPv4, line 32:
```suggestion
    "lint": "run-p lint:*",
```
Using `run-p` here makes it clearer they're running in parallel (npm-run-all's default behaviour)

package.json
```diff
@@ -27,18 +27,14 @@
...
+    "format:eslint": "npm run lint:eslint -- --fix",
+    "format:prettier": "npm run lint:prettier -- --write",
+    "format": "npm-run-all format:*",
```
MattIPv4, line 35:
```suggestion
    "format": "run-p format:*",
```

package.json
```diff
@@ -62,11 +58,28 @@
...
     "react-app-rewired": "^2.2.1",
     "typescript": "^4.9.5",
     "webpack-image-resize-loader": "^5.0.0"
+  },
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx,json}": [
+      "eslint src --ext .ts,.tsx",
+      "prettier --check \"src/**/*.{js,jsx,ts,tsx,json}\""
+    ]
+  },
```
MattIPv4, line 79: I would probably put this in a dedicated config file, to be consistent with how we're defining configs for other tools (e.g. eslint)

package.json
```diff
@@ -62,11 +58,28 @@
...
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
```
MattIPv4, line 83: Is this needed? The husky docs seem to suggest you just need the pre-commit file?

package.json
```diff
@@ -27,18 +27,14 @@
     "start": "react-app-rewired start",
     "build": "react-app-rewired build",
     "test": "react-app-rewired test",
-    "lint": "npm run lint:editorconfig && npm run lint:eslint",
-    "lint:editorconfig": "editorconfig-checker",
-    "lint:eslint": "eslint --ext .js,.jsx,.ts,.tsx src",
-    "lint:eslint:fix": "npm run lint:eslint -- --fix",
+    "lint:eslint": "eslint src --ext .ts,.tsx",
+    "lint:prettier": "prettier --check \"src/**/*.{js,jsx,ts,tsx,json}\"",
```
MattIPv4, line 31: We can probably also run prettier against the scss files?

.lintstagedrc.json
```diff
@@ -1,6 +1,7 @@
 {
-  "*.{js,jsx,ts,tsx,json,scss}": [
+  "*.{js,jsx,ts,tsx,json}": [
     "eslint src --ext .ts,.tsx",
-    "prettier --check \"src/**/*.{js,jsx,ts,tsx,json,scss}\""
-  ]
+    "prettier --check \"src/**/*.{js,jsx,ts,tsx,json}\""
+  ],
+  "*.{css,scss}": ["prettier --check \"src/**/*.{css,scss}\""]
```
jibranabsarulislam, line 6: prettier handles linting and formatting of scss fine, but i saw mentions in most guides talking about [stylelint](https://stylelint.io/user-guide/cli). Is this something that'd be worthwhile adding if prettier is already taking care of linting + formatting?

package.json
pjeweb: I think keeping `eslint-prettier-config` and `"extends": […, "prettier"]` is good practice to prevent eslint and prettier fighting or doing the same work twice. i would just keep it in there.

MattIPv4: :eyes: This comment was on the `eslint-config-react-app` line, not `eslint-config-prettier`

.lintstagedrc.json
MattIPv4: I think I'd consider it out of scope for this PR, but happy to accept a subsequent PR that adds stylelint following their recommended standards for SCSS

pjeweb: Prettier and stylelint check different things, so it might be interesting. I tried stylelint right now. There are a few syntax changes for example using the newer space separated color functions `rgba(0 0 255 / 80%)` instead of `rgba(0, 0, 255, 0.8)` etc. that i would call "nice-to-have". it also warns a lot about using CamelCase for identifiers instead of kebab-case. That would need to be refactored or configured. Imho the amount of scss in this repo is so small we dont really need a linter, but it would not hurt.

package.json
pjeweb: oops, yeah that can probably go

jibranabsarulislam: ah got it, my mistake 😅

MattIPv4:
👀 Looks like there are a couple of conflicts, and then this should be good to go

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged -c .lintstagedrc.json
```
MattIPv4, line 4:
```suggestion
npx lint-staged
```
I don't believe we need to specify the config here -- you're using a default filename for it, so it will automatically discover it

.lintstagedrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "*.{js,jsx,ts,tsx,json}": [
+    "eslint src --ext .ts,.tsx",
+    "prettier --check \"src/**/*.{js,jsx,ts,tsx,json}\""
```
MattIPv4, line 4:
```suggestion
    "eslint",
    "prettier --check"
```
You don't want to specify files here, that would defeat the whole point of lint-staged. lint-staged will automatically pass the specific files that have been modified and need checking

.lintstagedrc.json
```diff
@@ -0,0 +1,7 @@
...
+  ],
+  "*.{css,scss}": ["prettier --check \"src/**/*.{css,scss}\""]
```
MattIPv4, line 6:
```suggestion
  "*.{css,scss}": ["prettier --check"]
```
Ditto

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged
```
MattIPv4, line 4:
```suggestion
npx lint-staged

```
newline eof

MattIPv4 (approved):
LGTM, tyvm

PR #110, opened 2023-06-16
Allow multiple privileged users

MattIPv4:
With the extension being available for anyone during collabs, probably would be useful to have some more accounts that always have access to help/debug


pjeweb (dismissed):
well keep it at this and see if more is better on the collab?!

pjeweb (approved):
important !

PR #112, opened 2023-06-17
Move scroll detection to hook, require user event

### LINKED ISSUE #111: Unset command ambassador after X minutes
I propose hiding the card after 5 minutes (or so) of sleeping after being triggered by a command. Reason is: I think its weird to hover the screen and get the full card of an ambassador that was shown an hour ago. I would prefer the extension overlay to be in its default state.

MattIPv4:
Fixes #111 

When a chat command is run, and the user doesn't interact with the page, the card shown should auto-dismiss after 10s. When the user next interacts, the extension should not be showing any card.

### DIFF
--- a/src/pages/overlay/hooks/useScrolling.tsx
+++ b/src/pages/overlay/hooks/useScrolling.tsx
@@ -1,0 +1,78 @@
+import { useCallback, useEffect, useRef, useState } from "react";
+
+// Stop considering the user to be scrolling 100ms after the last scroll event
+const activityTimeout = 100;
+
+const useScrolling = () => {
+  // Allow subscriptions to scroll events
+  const [callbacks, setCallbacks] = useState<((e: Event) => void)[]>([]);
+  const on = useCallback((fn: (e: Event) => void) => {
+    setCallbacks((prev) => [...prev, fn]);
+  }, []);
+  const off = useCallback((fn: (e: Event) => void) => {
+    setCallbacks((prev) => prev.filter((f) => f !== fn));
+  }, []);
+
+  // Listen for the wheel event (user scrolling on desktop)
+  const onWheelTimer = useRef<NodeJS.Timeout | null>(null);
+  const onWheelActive = useRef(false);
+  const onWheel = useCallback(() => {
+    onWheelActive.current = true;
+
+    if (onWheelTimer.current) clearTimeout(onWheelTimer.current);
+    onWheelTimer.current = setTimeout(() => {
+      onWheelActive.current = false;
+    }, activityTimeout);
+  }, []);
+
+  // Listen for touch move events (user scrolling on mobile)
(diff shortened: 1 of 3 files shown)


src/pages/overlay/hooks/useScrolling.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const onTouchMoveTimer = useRef<NodeJS.Timeout | null>(null);
+  const onTouchMoveActive = useRef(false);
+  const onTouchMove = useCallback(() => {
+    onTouchMoveActive.current = true;
+
+    if (onTouchMoveTimer.current) clearTimeout(onTouchMoveTimer.current);
+    onTouchMoveTimer.current = setTimeout(() => {
+      onTouchMoveActive.current = false;
+    }, activityTimeout);
+  }, []);
```
MattIPv4, line 38: Open to other suggestions on how to ensure the scroll event came from a user interaction.

This isn't perfect, keyboard navigation isn't captured here for example -- I suppose we could also listen for any key press and treat scrolls coming from that as from the user too?

src/pages/overlay/hooks/useScrolling.tsx
```diff
@@ -0,0 +1,78 @@
...
+      node.addEventListener("wheel", onWheel, true);
+      node.addEventListener("touchmove", onTouchMove, true);
+
+      return () => {
+        node.removeEventListener("scroll", onScroll, true);
+        node.removeEventListener("wheel", onWheel, true);
+        node.removeEventListener("touchmove", onTouchMove, true);
+      };
+    },
+    [onScroll, onWheel, onTouchMove]
```
pjeweb, line 72: The ref function does not have a cleanup function like effects. So we might want to manually keep a ref to cleanup or just not cleanup=!

```suggestion
  const nodeRef = useRef<HTMLElement | null>(null);
  const ref = useCallback(
    (node: HTMLElement | null) => {
      if (!node) return;

      // cleanup old node
      if (nodeRef.current) {
        nodeRef.current.removeEventListener("scroll", onScroll, true);
        nodeRef.current.removeEventListener("wheel", onWheel, true);
        nodeRef.current.removeEventListener("touchmove", onTouchMove, true);
      }

      nodeRef.current = node;
      nodeRef.current.addEventListener("scroll", onScroll, true);
      nodeRef.current.addEventListener("wheel", onWheel, true);
      nodeRef.current.addEventListener("touchmove", onTouchMove, true);
    },
    [onScroll, onWheel, onTouchMove]
```

MattIPv4: Oh true, I had forgotten I'd written this as a ref -- I don't think what you've got here would work either (`onScroll` etc. could have changed identity). I'll fix this and push up something shortly 👍

pjeweb (approved):

PR #113, opened 2023-06-18
Fix ambassador card not reopening on repeated commands

pjeweb:
Bug repro:

- as a mod use chat command to open ambassador card (e.g. `!nilla`)
- as a viewer de-select the ambassador or select a different ambassador
- as a mod use chat command to open the same ambassador card (e.g. `!nilla`)
- expected: the ambassador of the chat command opens, actual: the overlay opens without select or the viewer selected card

### DIFF
--- a/src/pages/overlay/components/overlay/Overlay.tsx
+++ b/src/pages/overlay/components/overlay/Overlay.tsx
@@ -4,6 +4,8 @@ import React, {
   useCallback,
   useState,
   useMemo,
+  type SetStateAction,
+  type Dispatch,
 } from "react";
 
 import Buttons from "../buttons/Buttons";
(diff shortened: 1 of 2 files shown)


src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -96,6 +96,7 @@ export default function Overlay() {
           if (timeoutRef.current) clearTimeout(timeoutRef.current);
           timeoutRef.current = setTimeout(() => {
             setVisibleOption(undefined);
+            setCommandAmbassador(undefined);
```
MattIPv4, line 99: This fix will only work when the user doesn't interact with the popup triggered by a mod. If the user does immediately interact, this timer gets cancelled and so this reset will never happen.

We probably need a useEffect that immediately sets `commandAmbassador` back to `undefined` when it is any other value? Or maybe a `setImmediate` that does this reset?

pjeweb:
New approach, moving the state up to Overlay and passing the setter through context to not need the effect. I also extended the state to keep track of wether the active ambassador was selected by command to differentiate when to scroll.

MattIPv4 (approved):

Issue #64, opened 2023-03-20
Move away from create-react-app


abdullahmorrison:
small quality of life improvement, but It might be a lot of work so I'm not sure if it's worth it.

I want this project to be using react from scratch. Testing feels much smoother when if you configure webpack yourself. Also, wouldn't require third-party packages.

PR #121, opened 2023-07-09
Update dependencies

MattIPv4:


MattIPv4:
I had to add an override in here as cra/react-scripts doesn't seem to officially support typescript 5: https://github.com/facebook/create-react-app/issues/13080#issuecomment-1495014468

Hopefully with #64 we can move away from cra/react-app-rewired and just roll our own webpack config.

pjeweb (approved):

PR #130, opened 2023-10-06
Replace settings switch toggles with checkboxes

### LINKED ISSUE #129: Replace settings toggles with checkboxes
Toggles can be confusing to users as it is not clear whether they are enabled or not. We should replace the toggles in the settings popup of the overlay with checkboxes, which make it far clearer if they are checked or not.

MattIPv4:
Resolves #129 by replacing the existing switch toggle styling with custom checkboxes.

This PR also includes some consistency cleanup for the outline style we apply to things on hover, and a consistent shadow across all the elements.


pjeweb (approved):
looks nice and works

PR #137, opened 2023-10-08
UI cleanup (improved ambassador buttons/cards/list, SVG icons)

MattIPv4:
- Full-width image on ambassador cards/buttons
- Misc. ambassador card clean up
- Better fade on ambassador list overlay
- Use SVG icons for more things, less rasters

image

image


MattIPv4:
My only concern with this is the ambassador name overlay on the image in the cards -- I've added a bit of logic to offset the images where we can, and when hovering on the image the name fades away, but it's still not perfect. That being said, I think it looks a lot better than having the name elsewhere.

src/components/ambassadorButton/ambassadorButton.module.scss
```diff
@@ -36,16 +34,17 @@
...
-  .name {
-    color: $primary-text;
-    font-size: 0.8rem;
-  }
-
-  .species {
-    font-size: 0.7rem;
-    color: $secondary-text;
-    line-height: 1.2; // slightly reduce line height if species is multi-line
+    padding: 0.75rem 0.25rem;
```
pjeweb, line 37:
```suggestion
    padding: 0.4rem 0.25rem 0.6rem;
```

src/components/ambassadorCard/ambassadorCard.module.scss
```diff
@@ -12,111 +12,99 @@
...
       transition: $transition;
-      transition-property: color;
+      transition-property: opacity, backdrop-filter;
+
+      .name {
+        font-size: 1.3rem;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        overflow: hidden;
+        padding: 0 0.5rem;
```
pjeweb, line 73:
```suggestion
        padding: 0.2rem 0.5rem 0;
```

src/components/ambassadorCard/ambassadorCard.module.scss
```diff
@@ -12,111 +12,99 @@
...
 
   .scrollable {
+    display: flex;
+    flex-direction: column;
+    gap: 0.25rem;
     flex: 1 1 auto;
     overflow-y: auto;
-    margin-right: 1px;
+    margin: 0.25rem 0;
+    padding: 0 0.5rem 0.5rem;
```
pjeweb, line 107:
```suggestion
    margin: 0;
    padding: 0.5rem 0.5rem 1rem;
```

I feel like the scrollable needs a bit more vertical padding at the top. also not sure why we had separate margin here?

src/components/ambassadorCard/ambassadorCard.module.scss
```diff
@@ -132,15 +120,51 @@
...
+    .compact {
+      display: flex;
+      flex-wrap: wrap;
+      gap: 0.25rem 1.5rem;
+    }
+
+    .info {
+      display: inline-flex;
+      align-items: center;
+      gap: 0.5rem;
```
pjeweb, line 150:
```suggestion
      gap: 0.5rem;
      color: lighten($primary-color, 20%);
```

maybe do the lighten once and put it in a var

src/components/ambassadorCard/ambassadorCard.module.scss
```diff
@@ -132,15 +120,51 @@
...
-.birthday::after {
-  content: url("../../assets/partyHat.png");
-  width: 4rem;
-  height: auto;
-  position: absolute;
-  left: 50%;
-  top: 0;
-  transform: translate(-50%, -90%);
+    h3 {
+      font-size: 0.9rem;
```
pjeweb, line 125:
```suggestion
      font-size: 0.9rem;
      color: lighten($primary-color, 20%);
```

I feel like adding some color to the headings makes it more readable

pjeweb (commented):
Love the cleaned up UI! Just a few tweaks i would suggest

src/components/ambassadorCard/ambassadorCard.module.scss
MattIPv4: Using margin for the vertical spacing so that it remains present as you scroll -- margin happens outside scroll, padding happens inside.

src/components/ambassadorCard/ambassadorCard.module.scss
MattIPv4: Doing this seems wrong to me? The name then becomes no longer vertically centered in the blur area?

pjeweb: for me it does visually. the font sits a bit high in the line

MattIPv4: Ah yeah, I see now, does make it look more aligned to an imaginary baseline

pjeweb (approved):

PR #138, opened 2023-10-08
Make card close button an actual button (+ focus/hover consistency)

MattIPv4:


pjeweb (commented):
Do you think its worth opening an issue for making the extension keyboard accessible? There are plenty of problems with focus etc.

pjeweb (approved):

MattIPv4:
> Do you think its worth opening an issue for making the extension keyboard accessible? There are plenty of problems with focus etc.

Yeah -- I think basic keyboard navigation works with tab + enter, but definitely would be good to take a full pass and make it as accessible as possible.

PR #142, opened 2023-10-16
Change to tooltip to fit mobile panel

### LINKED ISSUE #115: Conservation tooltip does not fit mobile panel
Maybe we can open it to the bottom and align right.

image

t3st1n9:
Resolves #115 

Wrote a condition that checks whether tooltip fits to the right, if not tooltip is shown on the top. Also changed the position of the triangle to the bottom of the tooltip for that case.

overflowing
notoverflowing

### DIFF
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -42,8 +42,16 @@ const Tooltip = (props: TooltipProps) => {
         setIsOverflowing(true);
         setPosition({
           top: rect.top - rect.height / 2 - tooltipRect.height,
-          left: rect.right - tooltipRect.width / 2 - 5,
+          left: window.innerWidth / 2 - tooltipRect.width / 2,
         });
+        // Calculate margin so triangle will be pointing to info icon
+        const calcTriangleMargin = `${
+          window.innerWidth / 2 - rect.right + 5
+        }px`;
+        document.documentElement.style.setProperty(
+          "--tooltip-margin-right",
+          calcTriangleMargin,
+        );
       } else {
         // Position tooltip to the left
         setIsOverflowing(false);
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -28,7 +28,7 @@ const Tooltip = (props: TooltipProps) => {
   const [show, setShow] = useState(false);
   const [position, setPosition] = useState({ top: 0, left: 0 });
   const [above, setAbove] = useState(false);
-  const [triangleMargin, setTriangleMargin] = useState("0px");
+  const [triangleMargin, setTriangleMargin] = useState<string>();
 
   // On hover, compute position of tooltip and show it
   const handleEnter = useCallback(
@@ -47,7 +47,7 @@ const Tooltip = (props: TooltipProps) => {
           left: window.innerWidth / 2 - tooltipRect.width / 2,
         });
         // Calculate margin so triangle will be pointing to info icon
-        setTriangleMargin(`${window.innerWidth / 2 - rect.right + 5}px`);
+        setTriangleMargin(`0 ${window.innerWidth / 2 - rect.right + 5}px 0 0`);
       } else {
         // Position tooltip to the left
         setAbove(false);
@@ -55,6 +55,8 @@ const Tooltip = (props: TooltipProps) => {
           top: rect.top + rect.height / 2 - tooltipRect.height / 2,
           left: rect.right + 10,
         });
+        // Have the triangle be in the center of the tooltip
+        setTriangleMargin("-5px 0 0");
       }
       setShow(true);
     },
@@ -102,7 +104,7 @@ const Tooltip = (props: TooltipProps) => {
       >
         <div
           className={above ? styles.triangleBottom : styles.triangleLeft}
-          style={{ marginRight: above ? triangleMargin : "0" }}
+          style={{ margin: triangleMargin }}
         />
         {text}
       </div>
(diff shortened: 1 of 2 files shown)


src/components/tooltip/tooltip.module.scss
```diff
@@ -12,7 +12,16 @@
...
 
-  .triangle {
+  .triangle_bottom {
+    border-color: $tooltip-background-color transparent transparent transparent;
+    border-style: solid;
+    border-width: 5px;
+    margin-top: -5px;
+    position: absolute;
+    top: 105%;
+    right: 43.6%;
```
MattIPv4, line 22: This seems fragile and relies on the specific length of the current title for conservation status -- could we compute this instead?

src/components/tooltip/tooltip.module.scss
```diff
@@ -12,7 +12,16 @@
...
   z-index: 1;
 
-  .triangle {
+  .triangle_bottom {
+    border-color: $tooltip-background-color transparent transparent transparent;
+    border-style: solid;
+    border-width: 5px;
+    margin-top: -5px;
+    position: absolute;
+    top: 105%;
```
MattIPv4, line 21:
```suggestion
    top: calc(100% + 5px);
```
We want to be the full height, plus the 5px that is the size of the triangle itself

MattIPv4 (changes requested):

src/components/tooltip/tooltip.module.scss
t3st1n9: I changed the value of right to 50% and added margin variable that is calculated based on the position of info icon.
tooltip_length_of_text

I also considered centering tooltip based on the info icon and having the triangle in the middle of tooltip instead of centering tooltip in the ambassador card and having the triangle offset by margin, as it is right now. Unsure if it looks better.
centered_to_icon

src/components/tooltip/tooltip.module.scss
t3st1n9: After looking into it more, I noticed there was a margin-top that I copied from triangle_left, that was adding the -5px (so it would be aligned with the middle of info icon). When I removed it top could be 100%.

src/components/tooltip/tooltip.module.scss
MattIPv4: I think the current approach you have is best, as we know we'll always be in-bounds of the viewport if we remain within the card

src/components/tooltip/Tooltip.tsx
```diff
@@ -19,14 +19,14 @@ interface TooltipProps {
   fontSize?: string;
   maxWidth?: string;
 }
-
```
MattIPv4, line 22: Can we revert this linebreak removal?

src/components/tooltip/Tooltip.tsx
```diff
@@ -37,10 +37,29 @@ const Tooltip = (props: TooltipProps) => {
...
+          left: window.innerWidth / 2 - tooltipRect.width / 2,
+        });
+        // Calculate margin so triangle will be pointing to info icon
+        const calcTriangleMargin = `${
+          window.innerWidth / 2 - rect.right + 5
+        }px`;
+        document.documentElement.style.setProperty(
+          "--tooltip-margin-right",
+          calcTriangleMargin,
+        );
```
MattIPv4, line 54: Can we apply this in the computed `style` below, rather than mutating the document?

src/components/tooltip/tooltip.module.scss
```diff
@@ -12,7 +12,16 @@
   width: max-content;
   z-index: 1;
 
-  .triangle {
+  .triangle_bottom {
```
MattIPv4, line 15:
```suggestion
  .triangleBottom {
```
Very much a nit, but I think other classes in this repo use camelCase for classes

src/components/tooltip/tooltip.module.scss
```diff
@@ -12,7 +12,16 @@
...
+    border-color: $tooltip-background-color transparent transparent transparent;
+    border-style: solid;
+    border-width: 5px;
+    position: absolute;
+    margin-right: var(--tooltip-margin-right);
+    top: 100%;
+    right: 50%;
+  }
```
MattIPv4, line 23:
```suggestion
  }

```
nit, linebreak between blocks

src/components/tooltip/tooltip.module.scss
```diff
@@ -12,7 +12,16 @@
...
+  .triangle_left {
```
MattIPv4, line 24:
```suggestion
  .triangleLeft {
```
As mentioned in other comment, nit, but classes use camelCase by convention in this repo

src/components/tooltip/Tooltip.tsx
```diff
@@ -19,14 +19,14 @@ interface TooltipProps {
...
 }
-
 const Tooltip = (props: TooltipProps) => {
   const { text, children, fontSize, maxWidth } = props;
 
   const id = useId();
   const tooltipRef = useRef<HTMLDivElement>(null);
   const [show, setShow] = useState(false);
   const [position, setPosition] = useState({ top: 0, left: 0 });
+  const [isOverflowing, setIsOverflowing] = useState(false);
```
MattIPv4, line 29:
```suggestion
  const [above, setAbove] = useState(false);
```
I think naming this `above` might make it clearer what it indicates -- that the tooltip is now positioned above the trigger, rather than to the side of it? Overflowing to me felt too ambiguous, in that it could still be overflowing while above the tooltip if it were still too wide etc.

MattIPv4 (commented):
This is looking great testing locally, just a few nitpicks

src/components/tooltip/Tooltip.tsx
t3st1n9: I've removed the document mutation, but I'm not certain if my new approach is exactly what you suggested.

MattIPv4 (approved):
LGTM, tyvm

src/components/tooltip/Tooltip.tsx
MattIPv4: Ty! Almost exactly what I was looking for -- I pushed a very minor tweak to also do the non-above margin with it as well for consistency

PR #143, opened 2023-10-16
Remove overlay list blur

MattIPv4:
Much as this looked really nice in testing, on the actual extension this caused the video of the stream itself to be blurred, _and_ the blur doesn't work while the translate animation occurs, so it looks really janky. Removing it and relying on just the fade, which still looks fine.


pjeweb (approved):
Still looks great

PR #144, opened 2023-10-30
Don't fullscreen the video when double-clicking components

### LINKED ISSUE #54: Double-clicking components fullscreen's the video
double-clicking parts of the extension causes the video to fullscreen. If you're skimming through parts of the extension by clicking it fast enough, it can be annoying to get fullscreened.

The overlay needs to override the double-click functionality if the cursor is on it.

t1m0thyj:
Fixes #54 

Since I am not a React expert and this is my first time developing a Twitch extension, I welcome any critiques 😁 

I can remove cd084dcc9a23a8016967889e16da67b821b23510 if desired, but I believe using a tool like `cross-env` is necessary to set env variables inline on Windows.

### DIFF
--- a/src/pages/overlay/components/overlay/Overlay.tsx
+++ b/src/pages/overlay/components/overlay/Overlay.tsx
@@ -135,13 +135,11 @@ export default function Overlay() {
     return () => removeSleepListener("wake", callback);
   }, [addSleepListener, removeSleepListener]);
 
-  // Handle body clicks, dismissing the overlay if the user clicks outside of it
-  const bodyClick = useCallback((e: MouseEvent) => {
+  const visibleUnderCursor = (e: MouseEvent) => {
     // Get all the elements under the mouse
     const elements = document.elementsFromPoint(e.clientX, e.clientY);
 
-    // For each element, if it has a background then we want to ignore the click
-    // If we reach the body, then break out of the loop and close the panels
+    // For each element, if it has a background then it is part of the overlay
     for (const element of elements) {
       if (element === document.body) break;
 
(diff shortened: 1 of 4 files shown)


src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -164,6 +164,35 @@ export default function Overlay() {
     return () => document.body.removeEventListener("click", bodyClick, true);
   }, [bodyClick]);
 
+  // Handle body double clicks, ignoring them inside of overlay elements
+  const bodyDblClick = useCallback((e: MouseEvent) => {
+    // Get all the elements under the mouse
```
MattIPv4, line 169: Can we abstract this logic for finding a visible element under the mouse, given both this handler and the regular click handler are doing almost identical things? Maybe a `visibleUnderCursor` method that turns the first visible element? And then both the handlers could just call that in an if, instead of duplicating the for loop etc.

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -150,11 +148,18 @@ export default function Overlay() {
         style.backgroundImage !== "none" ||
         style.backgroundColor !== "rgba(0, 0, 0, 0)"
       ) {
-        return;
+        return true;
```
MattIPv4, line 151: Just to improve the utility here, I'd return the element here, and otherwise return nothing. Your if statements below won't change, but just makes this method slightly more flexible in the future.

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -135,13 +135,11 @@ export default function Overlay() {
     return () => removeSleepListener("wake", callback);
   }, [addSleepListener, removeSleepListener]);
 
-  // Handle body clicks, dismissing the overlay if the user clicks outside of it
-  const bodyClick = useCallback((e: MouseEvent) => {
+  const visibleUnderCursor = (e: MouseEvent) => {
```
MattIPv4, line 138: Could we move this into a util? Perhaps `utils/dom.ts`? Where it is currently, it will change identity every time React re-renders, and following the rules of hooks it should then be a dependency of your `useCallback`s, which'd then mean they're recomptued every re-render as well.

t1m0thyj: Thanks for the feedback 🙂 I've updated the method to return the element and moved it into a separate module.

MattIPv4 (approved):
LGTM, tyvm

PR #153, opened 2024-01-31
Default ambassador menu open, persist menu state

### LINKED ISSUE #134: Expand ambassador menu on first load, store menu state
To increase awareness of the extension and the ambassador information within, on first load of the extension for a user we should consider defaulting to the ambassador menu being shown.

We could then write to local storage what menu the user has opened, so if they close the menu and reload the page, it won't re-open on load. Similarly though, if they leave it open and reload the page, we'll re-open it on load.

trimbe:
Adds a setting to preserve menu state between loads. Defaults to `ambassadors` so all new users will see the `ambassadors` menu expanded on first load.

Closes #134


src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -19,23 +19,45 @@ const settings = {
...
+type SettingsProcessReturnTypes = {
+  boolean: boolean;
+  string: string;
 };
 
 type SettingsKey = keyof typeof settings;
 
 type StoredSettings = {
-  [key in SettingsKey]: ReturnType<(typeof settings)[key]["process"]>;
+  [key in SettingsKey]: key extends keyof SettingsProcessReturnTypes
```
MattIPv4, line 49: This doesn't seem right? `keyof SettingsProcessReturnTypes` is going to be `boolean | string`, but `key in SettingsKey` is going to be `"disableChatPopup" | "disableOverlayHiding" | "openedMenu"`?

src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -19,23 +19,45 @@ const settings = {
...
 };
 
 export type Settings = {
-  [key in SettingsKey]: (typeof settings)[key] & {
+  [key in SettingsKey]: {
+    title: string;
+    type: string;
+    process: (value: any) => boolean | string;
+    devOnly: boolean;
+    configurable: boolean;
```
MattIPv4, line 60: I think we can probably condense these down into a single `configurable` flag, where dev-only settings just do the environment comparison for that flag?

src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -19,23 +19,45 @@ const settings = {
...
+    ? SettingsProcessReturnTypes[key]
+    : any;
 };
 
 export type Settings = {
-  [key in SettingsKey]: (typeof settings)[key] & {
+  [key in SettingsKey]: {
+    title: string;
+    type: string;
+    process: (value: any) => boolean | string;
```
MattIPv4, line 58: Should this not be based on `SettingsProcessReturnTypes` so everything is tied to a single source of truth for the data types?

src/pages/overlay/components/overlay/settings/Settings.tsx
```diff
@@ -22,7 +22,7 @@ export default function Settings(props: OverlayOptionProps) {
 
           return (
             <li key={key}>
-              {setting.type === "boolean" && (
+              {setting.type === "boolean" && setting.configurable && (
```
MattIPv4, line 25: Instead of adding this check here, can we do it before the return, like the dev-only check?

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -89,6 +89,16 @@ export default function Overlay() {
...
 
+  // update setting when opened menu changes
+  useEffect(() => {
+    settings.openedMenu.change(visibleOption);
+  }, [visibleOption]);
+
+  // open saved (or default) menu when mounted
+  useEffect(() => {
+    setVisibleOption(settings.openedMenu.value);
+  }, []);
```
MattIPv4, line 100:
```suggestion
  }, [settings.openedMenu.value]);
```
It should be safe to have `settings.openedMenu.value` as a dependency here, to keep eslint happy?

MattIPv4 (changes requested):
Thoughts on having it be able to persist a specific ambassador card being open? Not required if not easy to do though.

src/pages/overlay/hooks/useSettings.tsx
trimbe: Sorry, I'm realizing now the typing changes I made are nonsense and unnecessary. Pulled most of it out, should be saner now.

src/pages/overlay/components/overlay/Overlay.tsx
trimbe: That would create a cycle since I'm updating the setting in the previous `useEffect` which depends on the state. I'm not seeing the `exhaustive-deps` rule in the project config though, but I can add an ignore if necessary.

trimbe:
> Thoughts on having it be able to persist a specific ambassador card being open? Not required if not easy to do though.

I don't think this would be hard but I'm not sure if it's useful really. As I see it, the main benefit here is just the default, persisting the menu is just a side effect. Up to you though.

src/pages/overlay/components/overlay/Overlay.tsx
MattIPv4: 🤔 Maybe I'm missing something, but it shouldn't create a cycle because they should be in sync? You set the setting to `a` in the previous effect, this effect runs and sets the visible option to `a`, at which point the previous effect won't run again because `a === a`?

MattIPv4:
👍 To save on scope creep, let's not worry about storing the specific card for now. Agree the main focus here is being open by default.

src/pages/overlay/components/overlay/Overlay.tsx
trimbe: Oh, I was initializing the state to an empty string, fixed.

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -59,7 +59,10 @@ const overlayOptions = [
   },
 ] as const;
 
-type OverlayKey = (typeof overlayOptions)[number]["key"];
+export const isValidOverlayKey = (key: string) =>
+  key == "" || overlayOptions.some((option) => option.key === key);
```
MattIPv4, line 63:
```suggestion
  key === "" || overlayOptions.some((option) => option.key === key);
```
Any reason we can't use strict equality here?

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -104,7 +119,8 @@ export default function Overlay() {
           // Dismiss the overlay after a delay
           if (timeoutRef.current) clearTimeout(timeoutRef.current);
           timeoutRef.current = setTimeout(() => {
-            setVisibleOption(undefined);
+            console.log(`setting visible option in timeout`);
```
MattIPv4, line 122:
```suggestion
```

src/pages/overlay/components/overlay/settings/Settings.tsx
```diff
@@ -17,16 +17,16 @@ export default function Settings(props: OverlayOptionProps) {
     <Card className={className} title="Extension Settings">
       <ul className={styles.settings}>
         {typeSafeObjectEntries(settings).map(([key, setting]) => {
-          if (setting.devOnly && process.env.NODE_ENV !== "development")
+          if (process.env.NODE_ENV !== "development" && !setting.configurable)
             return null;
```
MattIPv4, line 21: Instead of having this here, I think it'd be cleaner to just update the specific DEV-only setting to have `configurable: process.env.NODE_ENV === "development"`?

src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -25,15 +25,15 @@ const settings = {
...
+    configurable: process.env.NODE_ENV === "development",
   },
   openedMenu: {
     title: "Menu that was last opened",
     type: "string",
     process: (value: any): OverlayKey => {
       return isValidOverlayKey(value) ? value : "ambassadors";
     },
-    configurable: false,
+    configurable: process.env.NODE_ENV === "development",
```
MattIPv4, line 36: I would think this should always be false?

MattIPv4 (approved):

MattIPv4:
Tyvm for working on this 💚 Shout if there is anything we can improve with the DX for the project

PR #155, opened 2024-02-05
Add extra channels to connect to when not in default

### LINKED ISSUE #152: [Idea]: Always connect to AlveusGG channel
As a way to provide Alveus mods with control over the extension during collaborations, without needing them all added as `REACT_APP_CHAT_COMMANDS_PRIVILEGED_USERS`, _always_ connect to the AlveusGG channel (perhaps `REACT_APP_EXTRA_CHANNEL_NAMES`) alongside whatever channel the extension is running it.

To avoid the extension on the main Alveus channel triggering when mods in AlveusGG are running commands for the collaboration, `REACT_APP_EXTRA_CHANNEL_NAMES` should only be used when the current channel is not in `REACT_APP_DEFAULT_CHANNEL_NAMES`.

I _think_ the implementation should be something like:

https://github.com/alveusgg/extension/blob/fc2b989f162d55d503e16474e21d3c58b9d41815/src/hooks/useChatCommand.ts#L53

⬇️ 

```ts
    () => [...testChannelNames, ...(channel ? [channel, ...(defaultChannelNames.includes(channel) ? [] : extraChannelNames)] : defaultChannelNames)],
```

MattIPv4:
Resolves #152

Make sure your `.env` is updated to test, and this will need to be tested with a localhost extension on Twitch itself, as it will only connect to the extra channels when it knows what Twitch channel it is on.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,2 +1,3 @@
 REACT_APP_CHAT_COMMANDS_PRIVILEGED_USERS=abdullahmorrison,mattipv4,pjeweb,alveusgg,spacevoyage
 REACT_APP_DEFAULT_CHANNEL_NAMES=maya,alveussanctuary
+REACT_APP_EXTRA_CHANNEL_NAMES=alveusgg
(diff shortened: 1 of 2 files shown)


pjeweb (approved):
finally tested it. works great

Issue #67, opened 2023-03-28
Make ambassador card shiny


abdullahmorrison:
### [Twitter Inspo](https://twitter.com/MengTo/status/1640440942531321861?t=LAmLhaTA6kkz8ZfK2uuLJQ&s=19)

seems fun

abdullahmorrison:
This will probably distract from the information. May not be worth it.

MattIPv4:
https://www.npmjs.com/package/react-parallax-tilt is probably a clean way to implement this. We can add a user-controlled setting to disable this, but I think it'd be a fun thing to add.

Aarohi-Sharma:
hi, is this issue still up for grabs.
@MattIPv4

MattIPv4:
Yes, go for it

Aarohi-Sharma:
Sure, will have a look at it and work on this.

MattIPv4:
👋 @Aarohi-Sharma any update on this? I'd like to let someone else work on this if you're not planning to get a PR opened soon.

QuentinVdr:
Hey, I try to make something if the feature is still wanted.

PR #161, opened 2024-04-21
Make ambassador cards shiny

farrenr:
Resolves [#67](https://github.com/alveusgg/extension/issues/67)

- Used [react-parallax-tilt](https://www.npmjs.com/package/react-parallax-tilt) to create tilt and shine effects on ambassador cards.
- Added respective user setting to disable card effects.

### DIFF
--- a/src/components/ambassadorCard/AmbassadorCard.tsx
+++ b/src/components/ambassadorCard/AmbassadorCard.tsx
@@ -53,6 +53,7 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
       glarePosition="bottom"
       scale={1.0}
       perspective={2500}
+      tiltReverse
       className={classes(
         styles.ambassadorCard,
         className,
--- a/src/components/ambassadorCard/AmbassadorCard.tsx
+++ b/src/components/ambassadorCard/AmbassadorCard.tsx
@@ -32,20 +32,25 @@ export interface AmbassadorCardProps {
   ambassador: AmbassadorType;
   onClose?: () => void;
   className?: string;
+  disableCardEffects?: boolean;
 }
 
 export default function AmbassadorCard(props: AmbassadorCardProps) {
-  const { ambassadorKey, ambassador, onClose, className } = props;
+  const { ambassadorKey, ambassador, onClose, className, disableCardEffects } =
+    props;
   const images = getAmbassadorImages(ambassadorKey);
   const mod =
     window?.Twitch?.ext?.viewer?.role === "broadcaster" ||
     window?.Twitch?.ext?.viewer?.role === "moderator";
+  const glareOpacity = disableCardEffects ? 0.0 : 0.5;
 
   return (
     <Tilt
-      glareEnable={true}
-      glareMaxOpacity={0.5}
+      tiltEnable={!disableCardEffects}
+      glareEnable={!disableCardEffects}
+      glareMaxOpacity={glareOpacity}
       glareBorderRadius="1rem"
+      glarePosition="bottom"
       scale={1.0}
       perspective={5000}
       className={classes(styles.ambassadorCard, className)}
--- a/src/components/ambassadorCard/AmbassadorCard.tsx
+++ b/src/components/ambassadorCard/AmbassadorCard.tsx
@@ -52,134 +52,130 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
       glareBorderRadius="1rem"
       glarePosition="bottom"
       scale={1.0}
-      perspective={5000}
-      className={classes(styles.ambassadorCard, className)}
+      perspective={2500}
+      className={classes(
+        styles.ambassadorCard,
+        className,
+        ambassador.birth && isBirthday(ambassador.birth) && styles.birthday,
+      )}
     >
-      <div
-        className={classes(
-          ambassador.birth && isBirthday(ambassador.birth) && styles.birthday,
+      <div className={styles.hero}>
+        <img
+          className={styles.img}
+          src={images[0].src}
+          alt={images[0].alt}
+          style={{ objectPosition: offsetPosition(images[0].position) }}
+        />
+
+        <div className={styles.overlay}>
+          {props.onClose && (
+            <button
+              className={styles.close}
+              onClick={onClose}
+              type="button"
+              aria-label="Close"
+            >
+              &times;
+            </button>
+          )}
+
+          <h2 className={styles.name} title={ambassador.name}>
+            {ambassador.name}
+          </h2>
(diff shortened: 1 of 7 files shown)


src/components/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -30,142 +32,156 @@ export interface AmbassadorCardProps {
...
-      )}
+    <Tilt
+      tiltEnable={!disableCardEffects}
+      glareEnable={!disableCardEffects}
+      glareMaxOpacity={glareOpacity}
+      glareBorderRadius="1rem"
+      glarePosition="bottom"
+      scale={1.0}
+      perspective={5000}
+      className={classes(styles.ambassadorCard, className)}
```
MattIPv4, line 56: Unfortunately something here sets a transform-origin that makes this perspective shift look rather funky. I wonder if we need to move that transform-origin (which is used for the slide-in/out animation) into a wrapper around the Tilt component?

src/components/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -30,142 +32,156 @@ export interface AmbassadorCardProps {
...
-              &times;
-            </button>
-          )}
-
-          <h2 className={styles.name} title={ambassador.name}>
-            {ambassador.name}
-          </h2>
+      <div
+        className={classes(
+          ambassador.birth && isBirthday(ambassador.birth) && styles.birthday,
```
MattIPv4, line 60: The birthday hat looks to be broken?

src/components/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -30,142 +32,156 @@ export interface AmbassadorCardProps {
...
-        ambassador.birth && isBirthday(ambassador.birth) && styles.birthday,
-      )}
+    <Tilt
+      tiltEnable={!disableCardEffects}
+      glareEnable={!disableCardEffects}
+      glareMaxOpacity={glareOpacity}
+      glareBorderRadius="1rem"
+      glarePosition="bottom"
+      scale={1.0}
+      perspective={5000}
```
MattIPv4, line 55:
```suggestion
      perspective={7500}
      tiltReverse
```
From playing around with the settings locally, this felt nicer to me?

src/components/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -30,142 +32,156 @@ export interface AmbassadorCardProps {
...
-          )}
-
-          <h2 className={styles.name} title={ambassador.name}>
-            {ambassador.name}
-          </h2>
+      <div
+        className={classes(
+          ambassador.birth && isBirthday(ambassador.birth) && styles.birthday,
+        )}
+      >
```
MattIPv4, line 62: It looks like this extra inner div has also broken scrolling on the content?

MattIPv4 (changes requested):
👋 Looking great, but there are a few issues I noticed:

src/components/ambassadorCard/AmbassadorCard.tsx
farrenr: Ah I wasn't sure what was happening here, thanks for the spot - I'll take a look!

src/components/ambassadorCard/AmbassadorCard.tsx
farrenr: Not sure what happened here, will take a look at this and the birthday hat

src/components/ambassadorCard/AmbassadorCard.tsx
farrenr: When investigating this, I trialled removing the transform-origin from the ambassador card class within the overlay scss module and this appears to achieve the desired result. The cards still slide in from left and now the tilt works correctly. Let me know what you think, as far as I could tell, removing the transform-origin hasn't had any unwanted side effects!

src/components/ambassadorCard/AmbassadorCard.tsx
farrenr: Agree! Change made

src/components/ambassadorCard/AmbassadorCard.tsx
farrenr: Inner div removed as we now have the Tilt component div.

src/components/ambassadorCard/AmbassadorCard.tsx
```diff
@@ -30,17 +32,27 @@ export interface AmbassadorCardProps {
...
   return (
-    <div
+    <Tilt
+      tiltEnable={!disableCardEffects}
+      glareEnable={!disableCardEffects}
+      glareMaxOpacity={glareOpacity}
+      glareBorderRadius="1rem"
+      glarePosition="bottom"
+      scale={1.0}
+      perspective={2500}
```
MattIPv4, line 55:
```suggestion
      perspective={2500}
      tiltReverse
```
Can we enable `tiltReverse`? It feels more natural to me?

src/components/ambassadorCard/AmbassadorCard.tsx
MattIPv4, line 1: The only remaining issue looks to be that the tooltip on conservation status is now showing in the middle of nowhere?

farrenr: Good spot! Let me take a look

MattIPv4: 👋 @farrenr checking in if you had a chance to look at this?

farrenr: Hey! Apologies on taking a while with this, I've been a bit stuck on the last issue with the tooltip but I'm looking at it today

pjeweb:
I think it looks very cool, but I think it could be a bit smoother. The problem with it resetting on the edges and flickering seems to be inherent to the library: https://imgur.com/6nkHLFe

I think with increasing the transition speed to like 5s or so it gets a lot smoother overall, especially when leaving the card as it does not "snap" back to quickly. 

I also think the glare could be less or the font contrast gets too low. Its already not the best contrast. Maybe reducing the opacity to 30% could work.

Another thing i noticed, we do have a checkbox to disable the effect. Is there any way to also disable it when `prefers-reduced-motion` is active? Not sure how to handle the two - maybe just make the checkbox readonly with a note ("disabled due to browser preferences") or hide the checkbox completely?

MattIPv4:
👋 Given the lack of movement on this, and the conflicts (mostly my doing, sorry), I'm going to close this for now. Please feel free to open a new PR with updates if/when you have the time.

PR #167, opened 2024-07-29
Update dependencies

pjeweb:
Update various dependencies. Holding back on eslint v9 since the typescript tooling for it is still not stable.


package.json
```diff
@@ -11,7 +11,7 @@
     "url": "https://github.com/alveusgg/extension/issues"
   },
   "scripts": {
-    "prepare": "husky install",
+    "prepare": "husky",
```
MattIPv4, line 14: With husky v9 the first two lines of .husky/pre-commit should be removed

MattIPv4 (dismissed):
Otherwise, LGTM

PR #177, opened 2024-09-21
Add confetti for birthday

### LINKED ISSUE #72: Make confetti pop up on an ambassador's birthday

MattIPv4:
Resolves #72


pjeweb (approved):
very nice

PR #183, opened 2024-10-23
Replace SCSS modules styling with Tailwind

### LINKED ISSUE #15: Replace SCSS modules with Tailwind
To have a more consistent tech stack across the AlveusGG projects, look to replace the SCSS modules usage here with Tailwind.

Reference the tailwind config in https://github.com/alveusgg/alveusgg/blob/main/apps/website/tailwind.config.ts, and perhaps think about how we could even abstract that into a shared package (like alveusgg/data), so our theme is consistent.

MattIPv4:
Resolves #15


package.json
```diff
@@ -22,7 +22,8 @@
     "format": "run-p format:*"
   },
   "dependencies": {
-    "@alveusgg/data": "github:alveusgg/data#0.39.0",
+    "@alveusgg/data": "github:alveusgg/data#MattIPv4/tailwind-extras",
```
MattIPv4, line 25: Switch to released version

src/components/tooltip/Tooltip.tsx
```diff
@@ -96,14 +96,19 @@ const Tooltip = (props: TooltipProps) => {
     <>
       {childrenWithProps}
       <div
-        className={styles.tooltip}
+        className="bg-black/50 shadow-lg backdrop-blur rounded-lg p-2 pointer-events-none fixed w-max z-10 transition-opacity "
```
MattIPv4, line 99:
```suggestion
        className="bg-black/50 shadow-lg backdrop-blur rounded-lg p-2 pointer-events-none fixed w-max z-10 transition-opacity"
```

src/pages/overlay/components/overlay/settings/Settings.tsx
```diff
@@ -6,21 +6,23 @@ import Card from "../../../../../components/card/Card";
 import Toggle from "../../toggle/Toggle";
 
 import type { OverlayOptionProps } from "../Overlay";
-
-import styles from "./settings.module.scss";
+import { classes } from "../../../../../utils/classes";
```
MattIPv4, line 9: Sort import

MattIPv4 (commented):
Prettier class sorting?

src/pages/overlay/components/overlay/ambassadors/Ambassadors.tsx
```diff
@@ -84,58 +92,74 @@ export default function Ambassadors(props: OverlayOptionProps) {
...
-          className={styles.ambassadorCard}
-        />
-      )}
+      {ambassadors.map(([key]) => (
+        <Transition show={activeAmbassador.key === key} key={key}>
+          <AmbassadorCard
+            key={key}
+            ambassador={key}
+            onClose={() => setActiveAmbassador({})}
+            className="z-0 col-start-2 row-start-1 origin-[center_left] self-center transition-[opacity,transform] data-[closed]:-translate-x-10 data-[closed]:opacity-0 data-[closed]:motion-reduce:translate-x-0"
```
pjeweb, line 159: Chrome and Firefox are a bit jumpy without `will-change` for me shifting the text and card height a tiny bid after the transition. I think this fixed that.

Also I'd probably up the transition duration a bid. Feels a lot smoother to me.

```suggestion
            className="z-0 col-start-2 row-start-1 origin-[center_left] self-center transition-[opacity,transform] data-[closed]:-translate-x-10 data-[closed]:opacity-0 data-[closed]:motion-reduce:translate-x-0 duration-500 will-change-transform"
```

pjeweb (approved):
Very nice

src/pages/overlay/components/overlay/ambassadors/Ambassadors.tsx
pjeweb: Maybe not 500ms. I have not checked what the other durations are set to.

pjeweb:
Hmm and the IUCN tooltip does not sit right for me.

src/pages/overlay/components/overlay/ambassadors/Ambassadors.tsx
MattIPv4: Ah yeh, the old transitions globally were 300ms, Tailwind will default to 150ms. Will bump that back up.

MattIPv4:
> Hmm and the IUCN tooltip does not sit right for me.

Oh, how so?

pjeweb:
> > Hmm and the IUCN tooltip does not sit right for me.
> 
> Oh, how so?

Apparently thats a side effect of will-change ...

MattIPv4:
Ah yeh, setting will-change creates a new stacking context, which'll change how absolute positioning works

pjeweb (approved):

Issue #185, opened 2024-10-27
Update to ESLint v9 w/ flat config


MattIPv4:
Probably best not to use the migration tool they provide, let's write a clean config from scratch. It needs the eslint recommended rules, the typescript-eslint recommended rules + parser, the react recommended rules + plugin, and the prettier plugin.

MattIPv4:
https://github.com/search?q=eslint-plugin-react+typescript-eslint+path%3Aeslint.config.mjs&type=code is probably a good reference point for how other folks are doing minimal configs correctly with v9.

arilloid:
Hello @MattIPv4,

I'd love to work on this issue! I have just one question: for "the great recommended rules", did you mean to say "react" instead of "great"?

Could you please assign this issue to me?

MattIPv4:
> I'd love to work on this issue! I have just one question: for "the great recommended rules", did you mean to say "react" instead of "great"?

Yup, looks like auto-correct got me

> Could you please assign this issue to me?

Sure!

arilloid:
Sorry for any inconvenience caused, but I realized I’m unable to complete this issue. I’m finding it’s more complex than I initially thought (I got confused while storming the documentation). I’ll go ahead and unassign myself so someone else can take it on.

PR #186, opened 2024-10-27
Update dependencies

MattIPv4:
Notably, this does not update to ESLint v9 (#185).


pjeweb (changes requested):
I think you can bump up typescript-eslint to v8 so the warning about the newer typescript version disappears when linting/formatting. Does not need eslint v9.

pjeweb (approved):

PR #194, opened 2024-12-04
Add ambassador species lifespans

flakey5:
Closes #192

PR #195, opened 2024-12-06
Update data (mark Nilla Wafer as retired)

MattIPv4:
cc https://github.com/alveusgg/data/pull/84


MattIPv4:
Pulled the lifespan fix for Winston in from #194

pjeweb (approved):

PR #204, opened 2024-12-11
Upgrade Eslint to v9

### LINKED ISSUE #185: Update to ESLint v9 w/ flat config

guilherme-andrade:
Hi! Came across the repo and decided to open this :) Upgrades Eslint to v9, maintaining the same configuration while using the new flat config syntax.

Resolves #185, closes #201


package.json
```diff
@@ -14,7 +15,7 @@
     "prepare": "husky",
     "dev": "cross-env NODE_ENV=development webpack serve --config src/webpack.config.ts",
     "build": "webpack --config src/webpack.config.ts",
-    "lint:eslint": "eslint src --ext .js,.jsx,.ts,.tsx",
+    "lint:eslint": "eslint src 'src/**/*.{js,jsx,ts,tsx,css,json,html}'",
```
MattIPv4, line 18: We seem to now be specifying `src` twice, is that correct?

MattIPv4, line 18: Why are we sending css/json/html through eslint?

MattIPv4 (changes requested):
Thanks for working on this.

Please could you remove the old JSON eslint config?

eslint.config.js
```diff
@@ -0,0 +1,32 @@
+import eslint from "@eslint/js";
+import tseslint from "typescript-eslint";
+import pluginReact from "eslint-plugin-react";
+import eslintPluginPrettierRecommended from "eslint-plugin-prettier/recommended";
+
+export default [
+  eslint.configs.recommended,
+  ...tseslint.configs.recommended,
```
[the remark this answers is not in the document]
MattIPv4: Looking at https://github.com/alveusgg/alveusgg/blob/main/apps/website/eslint.config.js, the spread is correct.

package.json
guilherme-andrade: Oh oops, no that isn't. Will fix

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "license": "SEE LICENSE IN LICENSE.md",
+  "type": "module",
```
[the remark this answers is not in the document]
guilherme-andrade: Fixed now, converted `eslint.config.js` to an ESM file `eslint.config.mjs`

eslint.config.mjs
```diff
@@ -0,0 +1,33 @@
+// @ts-check
+import eslint from "@eslint/js";
+import tseslint from "typescript-eslint";
+import pluginReact from "eslint-plugin-react";
+import eslintPluginPrettierRecommended from "eslint-plugin-prettier/recommended";
+
+export default [
```
MattIPv4, line 7: Could you use `tseslint.config` here, like we are in https://github.com/alveusgg/alveusgg/blob/main/apps/website/eslint.config.js please (it will give us proper type-checking).

eslint.config.mjs
```diff
@@ -0,0 +1,33 @@
+// @ts-check
+import eslint from "@eslint/js";
+import tseslint from "typescript-eslint";
+import pluginReact from "eslint-plugin-react";
+import eslintPluginPrettierRecommended from "eslint-plugin-prettier/recommended";
```
MattIPv4, line 5: In https://github.com/alveusgg/alveusgg/blob/main/apps/website/eslint.config.js we're using eslint-config-prettier instead of just the plugin -- should we be using that here as well?

eslint.config.mjs
```diff
@@ -0,0 +1,33 @@
...
+
+export default [
+  eslint.configs.recommended,
+  ...tseslint.configs.recommended,
+  pluginReact.configs.flat?.recommended,
```
MattIPv4, line 10: I don't like the optional chain here -- I suspect we'll just want a ts-expect-error like we have in https://github.com/alveusgg/alveusgg/blob/main/apps/website/eslint.config.js?

package.json
```diff
@@ -42,7 +42,7 @@
     "cross-env": "^7.0.3",
     "css-loader": "^7.1.2",
     "dotenv-webpack": "^8.1.0",
-    "eslint": "^8.57.1",
+    "eslint": "^9.16.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-prettier": "^5.2.1",
```
MattIPv4, line 47: Related to my other note about prettier, I suspect we just need the config dep and not the plugin itself?

guilherme-andrade:
@MattIPv4 I couldn't pick this up again today, but will look into it this ASAP. I'll use the main website as basis for the next iteration.

MattIPv4:
No worries, and thank you!

MattIPv4 (approved):
LGTM, tyvm

MattIPv4:
FYI if you're interested, alveusgg/data also has an eslint config that needs upgrading ;)

PR #211, opened 2024-12-20
Update to Node.js 22

MattIPv4:
No longer pinning to a specific version, just anything 22.x, matching what we're doing in the site/data for consistency.


pjeweb (approved):

pjeweb (approved):
+1

PR #237, opened 2025-02-14
Migrate to Tailwind v4

### LINKED ISSUE #227: Migrate to Tailwind v4
Tailwind config needs to be migrated to v4 as CSS files. Investigate whether the scrollbar plugin supports v4 as a CSS plugin, or if it needs to be imported as a legacy JS plugin still (I don't think Tailwind has this functionality built-in still).

MattIPv4:
Resolves #227. cc https://github.com/alveusgg/data/pull/113.


pjeweb (approved):
nice work

Issue #256, opened 2025-03-13
Expand ambassador images on hover


abdullahmorrison:
Would be cool if I could view the whole image of the ambassadors when I hover my cursor over it. could be similar to how the name of the ambassador fades away when the image is hovered, part of the description could fade. 

A different solution could be adding an expand button on the bottom right of the image to show it full screen.

![Image](https://github.com/user-attachments/assets/f22bdc18-7374-4916-a993-11bcc29593b9)

MattIPv4:
+1 yeh I think it'd make sense to show the full image on hover somehow.

The images are meant to be better aligned -- it seems to work better in Firefox and at larger screen sizes. Probably need to revisit the CSS there again as part of this.

mirchez:
Hello! I'm interested in working on this issue. Could you please assign it to me?

I can implement the hover effect to expand the image or add a button for full-screen viewing, as discussed. I can also take a look at the alignment issues too.

mirchez:
Alright, perfect! I'll see what I can do.

MattIPv4:
#257

PR #257, opened 2025-03-15
Expand ambassador image on card hover

mirchez:
# Feat: Improve image responsiveness and interaction

##  Changes:
1. **Removed `overflow-hidden`** so images can have a minimum height when users zoom in on the page.  
2. **Added `rounded` to the image and ambassador name** to maintain a consistent layout.  
3. **Webpage:** Implemented `hover` pseudo-class to expand the image when the cursor is over it.  
4. **Mobile:** Implemented `active` pseudo-class to expand images when users tap on them.  

## Concerns  
- Since images have different widths and heights, some may appear zoomed in when expanded.  

## Before & After Screenshots  
extensionBefore
---------------------------------------
extensionAfter

---

**Related to #256**


src/components/AmbassadorCard.tsx
```diff
@@ -144,9 +144,9 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
             className="absolute top-0 left-1/2 z-10 h-auto w-16 -translate-x-1/2 -translate-y-[85%]"
           />
         )}
-        <div className="relative w-full overflow-hidden rounded-t-lg">
+        <div className="relative w-full rounded-t-lg">
           <img
-            className="peer aspect-[2.2] w-full object-cover sm:aspect-[1.8]"
+            className="peer aspect-[2.2] h-30 w-full rounded-t-lg object-cover transition-all duration-400 ease-in-out active:h-60 sm:aspect-[1.8] sm:h-32 sm:hover:h-64"
```
MattIPv4, line 149: 1. If you're setting both a width and a height here, aspect is no longer needed

2. When hovered can we use the intrinsic height instead? This will then avoid your concern about not all images being the same size

3. Nit, but can we just transition height, not everything?

.env.development.sample
```diff
@@ -1,3 +1,3 @@
 REACT_APP_TEST_CHANNEL_NAMES=abdullahmorrison
 REACT_APP_CHAT_COMMANDS_PRIVILEGED_USERS=abdullahmorrison
-REACT_APP_API_BASE_URL=http://localhost:3000
+REACT_APP_API_BASE_URL=http://localhost:3000
```
MattIPv4, line 3: Can you revert your change here?

MattIPv4 (commented):
Did you take a look at what's happening with the object position? I suspect that's part of the problem as I'm not sure it's consistent across browsers/viewports?

mirchez:
No problem, I'll work on it.

src/components/AmbassadorCard.tsx
mirchez: 1. Removed aspec.

2. Added transition only to height.

3. Transition doesn't work with intrinsic height and the expand it's very abrupt. 

I want to know if you want me to only use the intrinsic height, or if it's okay to set a certain height as long as the zoom doesn't interfere with the user experience.

.env.development.sample
mirchez: Done.

mirchez:
I did some changes

mirchez:
## Add max-height and fix objectPosition bug  

### Changes  
1. **Add `max-height`** to fix zoom when hovering; now the ambassador images can expand to their intrinsic height.  
2. **Set a default value for `objectPosition`** to prevent issues when the value is `undefined`, solving the alignment bug from before.  
3. **Apply transition only to `max-height`** for smoother expansion effects.  
4. **Remove `aspect` property**, as it is no longer needed.  

### Related to #256

src/components/AmbassadorCard.tsx
```diff
@@ -144,18 +144,23 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
...
-            className="peer aspect-[2.2] w-full object-cover sm:aspect-[1.8]"
+            className="peer max-h-30 w-full rounded-t-lg object-cover transition-[max-height] duration-700 ease-in-out active:max-h-96 sm:max-h-32 sm:hover:max-h-96"
             src={ambassador.image.src}
             alt={ambassador.image.alt}
             style={{
-              objectPosition: offsetPosition(ambassador.image.position),
+              objectPosition: offsetPosition(
+                ambassador.image.position
+                  ? ambassador.image.position
+                  : "50% 75%",
```
MattIPv4, line 157: I'm not sure I follow what this solves for -- `offsetPosition` already handles applying a default. I'm pretty sure the alignment issues stem from the way the min is applying itself currently within `offsetPosition` (the aim being to offset the images slightly beyond their normal positions to account for the title overlay)

mirchez: I see what you mean, I'm going to delete that part, it's not necessary.

MattIPv4:
Have you had a chance to investigate the alignment issues there were mentioned? I'm wondering if there's a better approach when can take than what happens in the `offsetPosition` method that's more consistent across viewports/browsers? Showing the full image on hover definitely partially solves for the issue, but also having the image aligned correctly when the card first shows is important too.

mirchez:
**`if (yPct <= 50) return `${x} min(calc(${y} + 1.5rem), 0%)`;`**

I quite don't understand why we use min here; it'll always return zero, isn't it?

MattIPv4:
So it shouldn't always be 0, the idea is that we shift the center of the image slightly so that it doesn't end up under the overlaid title -- I suspect this isn't always working as we expect. If you can't figure out what's wrong with it, it might be best we just use the normal center of the images without the 1.5rem offset, especially now that the full image shows on hover.

mirchez:
Agreed, I was trying to add some extra calculation, but I think it’s difficult to align because some ambassador images are larger than others. Now It just returns the normal center of the images, and it seems to align a little better. I'll commit it so you can take a look.

src/components/AmbassadorCard.tsx
```diff
@@ -22,9 +22,6 @@ const offsetPosition = (position?: `${number}% ${number}%`) => {
     `${number}%`,
   ];
 
-  const yPct = Number(y.replace("%", ""));
-  if (yPct <= 50) return `${x} min(calc(${y} + 1.5rem), 0%)`;
-
   return `${x} ${y}`;
```
MattIPv4, line 25: This is now just a no-op -- we can just remove the call to `offsetPosition` and apply `ambassador.image.position` directly

mirchez: Alright, as some ambassador.image.position will return undefined we should add a default value for **OjectPosition** as I did before, like this maybe:

```
objectPosition: offsetPosition(
                ambassador.image.position
                  ? ambassador.image.position
                  : "50% 50%",
               )
```

mirchez: Alright I'll remove it. as some of the ambassador.image.position will return undefined, we should add a default value for objectPosition, maybe like this: 

 style={{
              objectPosition: ambassador.image.position
                  ? ambassador.image.position
                  : "50% 50%" 
            }}

MattIPv4: `undefined` is fine, it'll just get dropped (and then the browser will use its default, which is to center it -- 50% 50%).

mirchez: Perfect, I'll commit the changes so you can review them.

MattIPv4:
Hm, with that calc/min offset removed we're definitely running into issues now where the set center of the image results in the ambassador's head being under the title:

image
image

If we can't figure out how to make that calc/min work more consistently, I think maybe its best if we don't have the title overlay? @pjeweb thoughts?

pjeweb:
Yeah I'd be just fine with moving the name to the main content.

mirchez:
What about something like this? 

image

MattIPv4:
I would maybe do it full width instead of having the margin, but yeh I think that's fine 👍

mirchez:
Here's how it would look with full width, but it seems to interfere with the scroll.


            <h2 className="-mx-2 -mt-2 mb-1 bg-alveus-green py-1 text-center text-base text-white transition-[scale] duration-300 ease-in-out hover:scale-102">
              {ambassador.name}
            </h2>


image

MattIPv4:
Can we put it outside the scroll container? If not the margin you had before is probably the better looking option

mirchez:
I think it looks better now.

image

MattIPv4:
👍 looks fine to me

src/components/AmbassadorCard.tsx
```diff
@@ -16,37 +16,18 @@ import partyHat from "../assets/party.svg";
...
-  return `${x} ${y}`;
-};
-
 const stringifyLifespan = (value: number | { min: number; max: number }) => {
   return typeof value === "number" ? `${value}` : `${value.min}-${value.max}`;
 };
 
 export interface AmbassadorCardProps {
   ambassador: string;
-  onClose?: () => void;
```
MattIPv4, line 37: I think we're going to need a mechanism for closing still -- on mobile the cards fill the entire screen so there is now no way to close them?

MattIPv4: Maybe we show a back arrow in the bar for the name, only when in the mobile/panel view?

mirchez: No problems, I'll work on it.

src/components/AmbassadorCard.tsx
```diff
@@ -48,7 +36,6 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
     ...extras
   } = props;
   const ambassador = useAmbassador(ambassadorKey);
-
```
MattIPv4, line 51: Nit, can we undo this unrelated line removal?

src/components/AmbassadorCard.tsx
```diff
@@ -144,36 +131,29 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
             className="absolute top-0 left-1/2 z-10 h-auto w-16 -translate-x-1/2 -translate-y-[85%]"
           />
         )}
-        <div className="relative w-full overflow-hidden rounded-t-lg">
+        <div className="relative w-full rounded-t-lg">
```
MattIPv4, line 134: Do we even need this div now?

src/components/AmbassadorCard.tsx
```diff
@@ -144,36 +131,29 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
...
           <img
-            className="peer aspect-[2.2] w-full object-cover sm:aspect-[1.8]"
+            className="peer max-h-30 w-full rounded-t-lg object-cover transition-[max-height] duration-700 ease-in-out active:max-h-96 sm:max-h-32 sm:hover:max-h-96"
```
MattIPv4, line 136: I don't think we need `peer` now as we're not doing any styling based on that?

src/components/AmbassadorCard.tsx
```diff
@@ -144,36 +131,29 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
...
-              <button
-                className="absolute top-1/2 right-1 block w-8 -translate-y-1/2 cursor-pointer text-center text-2xl transition-colors hover:text-highlight focus:text-highlight"
-                onClick={onClose}
-                type="button"
-                aria-label="Close"
-              >
-                &times;
-              </button>
-            )}
+        <div className="flex w-full items-center justify-center bg-alveus-green py-1">
```
MattIPv4, line 146: I think we want relative here so the arrow is positioned relative to here?

src/components/AmbassadorCard.tsx
```diff
@@ -144,36 +131,29 @@ export default function AmbassadorCard(props: AmbassadorCardProps) {
...
+          <button
+            className="absolute w-8 cursor-pointer text-2xl text-alveus-green-200 transition-colors hover:text-highlight focus:text-highlight sm:hidden"
```
MattIPv4, line 148: I think we need to set `left-0` on this?

src/components/AmbassadorCard.tsx
MattIPv4, line 146: Also, we should have a `px-?` set so that the ambassador name can never intersect with the arrow, please

src/components/AmbassadorCard.tsx
MattIPv4: Thank you 💚

mirchez:
I've added a new Icon here:

image

So the arrow now looks like this:

image

Is it okay?

MattIPv4:
I'd maybe make the arrow slightly lighter if we can, and as a super nit, can the px-15 be px-16? Otherwise, in principle, that looks great -- thank you!

mirchez:
Perfect, I'll do those changes and then commit.

MattIPv4:
👍 Looks great

MattIPv4 (approved):
Thank you very much for your work on this 💚

mirchez:
Thanks Matt for the patience. I really appreciate it. I wish you all the best.

PR #332, opened 2025-07-10
Include "Did you know?" when fact is present and show on 'back' of AmbassadorCard

itsjayway:
Reference: https://github.com/alveusgg/extension/issues/325

Hey all, I'm back! This is an MVP of implementation. It's currently missing the tailwind 'flip' animation as of the current commits and I'll continue toward delivering that in this PR.

If this is sufficient to proceed, I'm looking for feedback in:

- [ ] "Did you know? ℹ️" placement on the 'front' of the card.
- [ ] Card 'back' formatting
- [ ] Implementation

https://github.com/user-attachments/assets/61e563e2-de38-4f56-a16a-77dc5d3d1abd

### DIFF
--- a/src/components/Dogear.tsx
+++ b/src/components/Dogear.tsx
@@ -21,6 +21,9 @@ export default function CardDogear(props: {
           className={classes(
             "absolute top-0 right-0 w-0 border-13 border-solid border-t-transparent border-r-transparent border-b-alveus-green border-l-alveus-green transition-all duration-300 ease-in-out hover:scale-[1.6] sm:border-15 md:border-18 lg:border-20",
           )}
+          style={{
+            boxShadow: "-4px 4px 2px rgba(0,0,0,0.3)",
+          }}
         />
       </button>
     </Tooltip>
--- a/src/components/Dogear.tsx
+++ b/src/components/Dogear.tsx
@@ -2,27 +2,6 @@ import React from "react";
 import Tooltip from "./Tooltip";
 import { classes } from "../utils/classes";
 
-/* 
-<button
-            className="absolute top-0 right-0 z-20 m-0 h-5 w-5 bg-transparent p-0"
-            style={{ outline: "none" }}
-            onClick={() => {
-              alert("Dogear clicked!");
-            }}
-            aria-label="Dogear"
-            type="button"
-          >
-            {/* Credits to https://codepen.io/rachelslurs/pen/xxoPKLg *\/
-            <div
-              className={classes(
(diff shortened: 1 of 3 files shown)


MattIPv4:
👀 Thank you for working on this! I think I agree with a lot of the feedback that copilot is giving.

Instead of having it be a heading with an icon that you need to clip, perhaps we could do a peeking/peeling corner that the user can click on to flip the card over? And perhaps when they hover on that corner we show a tooltip to give context on clicking to flip?

itsjayway:
Currently the simplest way I could manage rendering the dogear. It should be possible to have the fold "bring more with it" via translate-x/y, but that would reveal the unclipped portion of the card. 

https://github.com/user-attachments/assets/6e5d0be4-5034-4e81-8c63-324815223751

MattIPv4:
That looks great -- I think with a shadow on that, and the tooltip, it should be just fine.

src/hooks/useAmbassadors.tsx
```diff
@@ -53,6 +53,7 @@ const apiAmbassadorSchema = ambassadorSchema.extend({
     key: z.string(),
     title: z.string(),
   }),
+  fact: z.string().optional(),
```
MattIPv4, line 56: Is the fact not already part of `ambassadorSchema`?

MattIPv4:
One other thought, I wonder if it'd be worth splitting `AmbassadorCard.tsx` into `ambassador/card-front.tsx` + `ambassador/card-back.tsx` + `ambassador/card.tsx` (and `AmbassadorButton.tsx` could also move to `ambassador/button.tsx`)?

MattIPv4:
@itsjayway how're things going, any progress update on here?

src/components/Dogear.tsx
```diff
@@ -0,0 +1,31 @@
...
+        className="absolute top-0 right-0 z-20 m-0 h-5 w-5 bg-transparent p-0"
+        style={{ outline: "none" }}
+        onClick={props.onClick}
+        aria-label="Dogear"
+        type="button"
+      >
+        {/* Credits to https://codepen.io/rachelslurs/pen/xxoPKLg */}
+        <div
+          className={classes(
+            "absolute top-0 right-0 w-0 border-13 border-solid border-t-transparent border-r-transparent border-b-alveus-green border-l-alveus-green transition-all duration-300 ease-in-out hover:scale-[1.6] sm:border-15 md:border-18 lg:border-20",
```
itsjayway, line 22: magic numbers 😩 ensures consistent sizing across window widths. an alternative could be `borderWidth: "clamp(13px, 2vw, 20px)"`, but it was not satisfactory for certain window widths regardless the preferred value

PR #362, opened 2025-09-20
Add tilt card to ambassador card

QuentinVdr:
Resolves [Make ambassador card shiny #67](https://github.com/alveusgg/extension/issues/67)

- Add tilt and glare effects on ambassador card
- Add disable settings


src/components/TiltCard.tsx
```diff
@@ -0,0 +1,142 @@
...
+  disabled?: boolean;
+}
+
+export default function ConditionalTiltCard({
+  children,
+  disabled = false,
+  maxTilt = 15,
+  glareMaxOpacity = 0.3,
+  className = "",
+}: Readonly<ConditionalTiltCardProps>) {
```
MattIPv4: Why `Readonly` here, not seen that before on prop typings?

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,142 @@
...
+function TiltCard({
+  children,
+  maxTilt = 15,
+  glareMaxOpacity = 0.3,
+  className = "",
+}: Readonly<TiltCardProps>) {
+  const cardRef = useRef<HTMLDivElement>(null);
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const [tiltStyle, setTiltStyle] = useState<React.CSSProperties>({});
+  const [glareStyle, setGlareStyle] = useState<React.CSSProperties>({});
```
MattIPv4: Please import `CSSProperties` rather than relying on `React` being global

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,142 @@
...
+        className="pointer-events-none absolute -inset-4"
+        onMouseMove={isHovered ? handleMouseMove : undefined}
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handleMouseLeave}
+        style={{ pointerEvents: isHovered ? "auto" : "none" }}
+      />
+
+      <div
+        ref={cardRef}
+        className={`relative overflow-visible will-change-transform ${className}`}
```
MattIPv4: Use the `classes` util for composing class names, please

MattIPv4 (commented):
👀 Thank you for working on this. From testing locally, a few thoughts:

1. The way the glare is applied doesn't feel particularly realistic in my testing, it does just look/feel like a radial gradient centered around whereever the mouse is, rather than a consistent light source bouncing off the card?

2. If you have a card open, and then click on an ambassador to switch cards, the transition animation doesn't appear to be working correctly anymore?

3. When the effect is enabled, the card is no longer vertically centered?

4. What would it look like to have the card tilt toward the cursor rather than away from it? I feel having it tilt away does make the UX that bit harder as when you go to click on a link in the card for example, it feels like the card is trying to move away from you.

5. Some consideration needs to be had for the mobile/panel views -- either we need to expose settings there, or probably default to this being disabled?

6. More generally, could you speak to the decision to implement this all by hand rather than relying on an existing package that'd do this for us without the maintenance burden of doing it ourselves?

src/components/TiltCard.tsx
QuentinVdr: The Readonly is a good practice i discover with sonarqube : https://next.sonarqube.com/sonarqube/coding_rules?open=typescript%3AS6759&rule_key=typescript%3AS6759

MattIPv4: Ah I see, I don't think that applies here really as you're doing inline destructing, so you never get a mutable `props` object anyway? I'd rather stick to the existing conventions in the repo and implement this separately across everything if deemed needed.

QuentinVdr:
6. I prefer to implement this by hand cause it wasn't looking that hard and the library you advice and other i saw got the flicking issues for the hover when the card is tilt. If you prefer i can try using the library you advice. It's as you want.
1. Yeah I try some stuff but i don't really know how to make the glare feel natural.
2-3. I just saw that gonna think to rework the tilt card to fix it. I think i know where it's come and it's the same origin for both of these issues.
4. Ok i was feeling that is more natural by going away from the mouse. But yeah make sens that can be annoying for UX.
5. Yeah that true i think just disable it for now is the best.

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+  forwardRef,
+  type CSSProperties,
+} from "react";
+import { classes } from "../utils/classes";
+
+interface ConditionalTiltCardProps extends TiltCardProps {
+  disabled?: boolean;
+}
+
+const ConditionalTiltCard = forwardRef<
```
MattIPv4: We're using React 19, so forwardRef shouldn't be needed?

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+  HTMLDivElement,
+  ConditionalTiltCardProps
+>(function ConditionalTiltCard(
+  {
+    children,
+    disabled = true,
+    maxTilt = 15,
+    glareMaxOpacity = 0.4,
+    className = "",
```
MattIPv4: Why apply defaults here, as well as in `TiltCard`? I would think we'd only want to define them in one place?

MattIPv4: `className` should be fine to default to `undefined`, rather than an empty string?

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+interface TiltCardProps {
+  children: React.ReactNode;
+  maxTilt?: number;
+  glareMaxOpacity?: number;
+  className?: string;
+}
+
+const TiltCard = forwardRef<HTMLDivElement, TiltCardProps>(function TiltCard(
+  { children, maxTilt = 15, glareMaxOpacity = 0.4, className = "", ...extras },
```
MattIPv4: `className` should be fine to default to `undefined`, rather than an empty string?

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+      className={className}
+      {...extras}
+    >
+      {children}
+    </TiltCard>
+  );
+});
+
+interface TiltCardProps {
+  children: React.ReactNode;
```
MattIPv4: Import the `ReactNode` type please, rather than relying on the global `React`

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+  ref,
+) {
+  const cardRef = useRef<HTMLDivElement>(null);
+
+  const callbackRef = useCallback(
+    (node: HTMLDivElement | null) => {
+      if (ref) {
+        if (typeof ref === "function") ref(node);
+        else ref.current = node;
+      }
```
MattIPv4, line 73: Why is this wrapped in `if (ref)`, won't this prevent `null` being propagated to the passed in ref?

src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -38,6 +38,12 @@ const settings = {
     },
     configurable: false,
   },
+  disableCardEffects: {
+    title: "Disable Ambassador Card Effects",
+    type: "boolean",
+    process: (value: any) => !!value,
+    configurable: true,
+  },
```
MattIPv4: Nit, this should probably go above `disableOverlayHiding`, so that the regular user-configurable settings are grouped together?

MattIPv4 (commented):
Glare looks quite a bit better now to my eye, nice job.

MattIPv4:
@pjeweb would you be able to also take a look through this?

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+  disabled?: boolean;
+}
+
+const ConditionalTiltCard = forwardRef<
+  HTMLDivElement,
+  ConditionalTiltCardProps
+>(function ConditionalTiltCard(
+  {
+    children,
+    disabled = true,
```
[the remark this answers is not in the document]
MattIPv4: Agree with this, the `disabled` prop should default to false, and you need to pass `disabled` if you want it disabled.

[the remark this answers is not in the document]
MattIPv4: Semi-related, it might be worth making the setting for this have an intelligent default if the user hasn't set one before, based on whether they've got reduced motion enabled or not?

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+      const rect = cardRef.current.getBoundingClientRect();
+      const centerX = rect.left + rect.width / 2;
+      const centerY = rect.top + rect.height / 2;
+
+      const mouseX = (e.clientX - centerX) / (rect.width / 2);
+      const mouseY = (e.clientY - centerY) / (rect.height / 2);
+
+      const clampedMouseX = Math.max(-1.5, Math.min(1.5, mouseX));
+      const clampedMouseY = Math.max(-1.5, Math.min(1.5, mouseY));
```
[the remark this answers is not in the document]
MattIPv4: Very obvious what these are, no need.

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+      const tiltX = clampedMouseY * maxTilt;
+      const tiltY = clampedMouseX * maxTilt * -1;
+
+      const lightSourceX = -0.3;
+      const lightSourceY = -0.5;
```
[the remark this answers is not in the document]
MattIPv4: No Copilot, these are already constants. _But_, I do think it might be good to have some comments explaining what these values are, and more generally, what the computations are doing in this logic?

src/components/TiltCard.tsx
QuentinVdr: I'm sorry but i don't understand that comment. Cause if there is no ref passed or it's null we can't set the node to it. Or there is something i don't know about and i don't understand ?

src/components/TiltCard.tsx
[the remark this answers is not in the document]
QuentinVdr: Are the comment clear or not ?

src/components/TiltCard.tsx
MattIPv4: Oh sorry, I misread this, I thought this was doing `if (node)` not `if (ref)`

src/components/TiltCard.tsx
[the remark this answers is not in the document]
MattIPv4: The comments now look great, thank you!

pjeweb:
So I love the idea and the glare looks great. I think its quite a lot though. Maybe we can reduce the tilt a bit? Also how hard would it be to make the tilt when entering from the side ease too? I you go over the card fairly quickly with the mouse, it is very wiggly, if that makes sense.

https://github.com/user-attachments/assets/c4299d50-a74e-45ff-8026-16333f9bac57

The video is like the worst case, I know, but just to illustrate ;)

Code looks clean though! Thanks for working on this!

QuentinVdr:
I make the animation smoother but i feel like it's slow to react on hover. I don't feel it's an issues. But if you don't like I can try to rethink and add a time interval to make the card smooth for the 0.3s after entering and leaving the card and the rest of the time just react on time.

pjeweb:
I think thats way better. Maybe you can increase max tilt again a bit, but either way.

@MattIPv4 happy to approve. you decide whats looking/feeling better :)

pjeweb (dismissed):

QuentinVdr:
I tweak the transition do get a smooth transition on entering the card with the mouse still moving but feel smooth to me

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,222 @@
...
+      }
+      cardRef.current = node;
+    },
+    [ref],
+  );
+
+  const [tiltStyle, setTiltStyle] = useState<CSSProperties>({});
+  const [glareStyle, setGlareStyle] = useState<CSSProperties>({});
+  const [isHovered, setIsHovered] = useState(false);
+  const [transitionTimeout, setTransitionTimeout] = useState<NodeJS.Timeout>();
```
MattIPv4: This should be a ref, not a state, so that you don't end up in a effect dependency loop

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,222 @@
...
+        "all 0.35s cubic-bezier(0.25, 0.46, 0.45, 0.94)";
+    }
+
+    if (transitionTimeout) clearTimeout(transitionTimeout);
+    const timeout = setTimeout(
+      () => {
+        if (cardRef.current) cardRef.current.style.transition = "none";
+      },
+      isHovered ? 400 : 350,
+    );
```
MattIPv4: Out of curiosity, is there a specific reason for the slightly different times for in/out, rather than just using the same duration for both?

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,222 @@
...
+    setTransitionTimeout(timeout);
+
+    return () => {
+      if (timeout) clearTimeout(timeout);
+    };
+  }, [isHovered]);
+
+  // Light reflection angle based on tilt
+  // Simulate light source coming from top-left (more natural)
+  const lightSource = { x: -0.3, y: -0.5 };
```
MattIPv4: This will want to be defined outside the component at the top of this file, otherwise it'll need to be a dependency of the callback below, and then it'd be unstable unless you memoize this object

src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -24,6 +24,13 @@ const settings = {
     process: (value: any) => !!value,
     configurable: true,
   },
+  disableCardEffects: {
+    title: "Disable Ambassador Card Effects",
+    type: "boolean",
+    process: (value: any) =>
+      value ?? window.matchMedia("(prefers-reduced-motion: reduce)").matches,
```
MattIPv4, line 31: I don't think this guarantees that the value will actually be a boolean? Could just wrap this entire expression inside `!!()`?

MattIPv4:
Oh, and it looks like the transition when a card closes doesn't work anymore again?

src/pages/overlay/hooks/useSettings.tsx
QuentinVdr: the `.matches` return a boolean but i can still add the `!!()` if you want

src/components/TiltCard.tsx
QuentinVdr: was just feeling more natural cause the transition function are different

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,209 @@
...
+      cardRef.current = node;
+    },
+    [ref],
+  );
+
+  const [tiltStyle, setTiltStyle] = useState<CSSProperties>({});
+  const [glareStyle, setGlareStyle] = useState<CSSProperties>({});
+  const [isHovered, setIsHovered] = useState(false);
+  const [tiltTransition, setTiltTransition] = useState("");
+  let transitionTimeout: NodeJS.Timeout;
```
MattIPv4: This needs to be a `useRef`, otherwise the value will not persist across re-renders.

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,209 @@
...
+    });
+  }, []);
+
+  const setTiltTransitionWithTimeout = (transition: string) => {
+    setTiltTransition(transition);
+    if (transitionTimeout) clearTimeout(transitionTimeout);
+    transitionTimeout = setTimeout(() => {
+      setTiltTransition("duration-0 ease-out");
+    }, 300);
+  };
```
MattIPv4: Wrap in `useCallback`, please.

src/components/TiltCard.tsx
```diff
@@ -0,0 +1,209 @@
...
+    setTiltTransitionWithTimeout("duration-300 ease-out");
+    setTiltStyle({
+      transform:
+        "perspective(1000px) rotateX(0deg) rotateY(0deg) scale3d(1, 1, 1)",
+    });
+    setGlareStyle({
+      background: "none",
+      opacity: 0,
+    });
+  }, []);
```
MattIPv4: These are both missing `setTiltTransitionWithTimeout` as a hook dependency

PR #371, opened 2025-10-02
Improve keyboard accessibility

### LINKED ISSUE #139: Make extension keyboard accessible
Make the extension work with keyboard input:
 - make keyboard focus visible
 - allow closing modals with esc
 - prevent tabbing outside of modals or into hidden elements
 - to not move or loose focus when interacting
 - buttons and other elements should follow common input patterns (space/enter, cursor keys, escape etc.)
 - probably forgot some points otomh

flakey5:
Closes #139

### DIFF
--- a/src/pages/overlay/App.tsx
+++ b/src/pages/overlay/App.tsx
@@ -19,6 +19,8 @@ export default function App() {
     sleeping,
     wake,
     sleep,
+    stayAwake,
+    resumeTimer,
     on: addSleepListener,
     off: removeSleepListener,
   } = useSleeping();
@@ -54,6 +56,8 @@ export default function App() {
         onTouchMove={interacted}
         onKeyDown={interacted}
         onMouseLeave={sleep}
+        onFocus={() => stayAwake()}
+        onBlur={() => resumeTimer(timeout)}
       >
         <Overlay />
       </div>
--- a/src/pages/overlay/components/overlay/Ambassadors.tsx
+++ b/src/pages/overlay/components/overlay/Ambassadors.tsx
@@ -4,6 +4,7 @@ import {
   useCallback,
   useMemo,
   type MouseEvent,
+  type KeyboardEventHandler,
 } from "react";
 import { Transition } from "@headlessui/react";
 
@@ -19,6 +20,8 @@ import type { OverlayOptionProps } from "./Overlay";
 
 import IconChevron from "../../../../components/icons/IconChevron";
 import useSettings from "../../hooks/useSettings";
+
+import * as keyBinds from "../../../../keyBinds";
 
 const arrowClass =
   "absolute border-0 cursor-pointer text-alveus-green w-full h-[var(--list-fade-padding)] z-20 transition-opacity group pt-[var(--twitch-vertical-padding)] pb-4 box-content";
@@ -50,6 +53,8 @@ export default function Ambassadors(props: AmbassadorsProps) {
     [rawAmbassadors, plants],
   );
 
+  const activeAmbassadorButtonRef = useRef<HTMLButtonElement>(null);
+  const activeAmbassadorRef = useRef<HTMLDivElement>(null);
   const upArrowRef = useRef<HTMLButtonElement>(null);
   const ambassadorList = useRef<HTMLDivElement>(null);
   const downArrowRef = useRef<HTMLButtonElement>(null);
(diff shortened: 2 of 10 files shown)


src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -233,6 +241,40 @@ export default function Overlay() {
...
+      return;
+    }
+  };
+
+  const activeOverlayRefCallback = useCallback(
+    (node: HTMLDivElement | null) => {
+      activeOverlayRef.current = node;
+
+      // Auto focus the overlay that's currently selected
+      node?.focus();
```
flakey5: This part doesn't seem to be working and I can't really tell why?

MattIPv4:
Sorry for the delay on getting to this @flakey5, been pretty busy with some upcoming secret bits for Alveus. Any chance you could rebase this?

src/utils/keyBinds.ts
MattIPv4, line 1: Can we put this in src/utils please?

src/pages/overlay/App.tsx
```diff
@@ -54,6 +56,8 @@ export default function App() {
         onTouchMove={interacted}
         onKeyDown={interacted}
         onMouseLeave={sleep}
+        onFocus={() => stayAwake()}
+        onBlur={() => resumeTimer(timeout)}
```
MattIPv4: Just want to confirm that this isn't going to interfere with the automatic sleeping, clicking on buttons in the UI won't cause onFocus to fire?

src/pages/overlay/components/overlay/Ambassadors.tsx
```diff
@@ -137,15 +193,19 @@ export default function Ambassadors(props: AmbassadorsProps) {
...
           onScroll={handleArrowVisibility}
+          tabIndex={-1}
         >
           {ambassadors.map(([key]) => (
             <AmbassadorButton
               key={key}
               ambassador={key}
-              onClick={() => {
+              onClick={(event) => {
+                // @ts-expect-error this is fine
```
MattIPv4: Can we avoid this expect-error?

src/pages/overlay/components/overlay/Overlay.tsx
MattIPv4: What is this meant to do that it isn't doing? Focus the entire active overlay (e.g. the entire ambassadors list)?

src/pages/overlay/components/overlay/Overlay.tsx
```diff
@@ -269,10 +311,16 @@ export default function Overlay() {
           ) &&
           hiddenClass,
       )}
+      onKeyDown={onKeyDown}
     >
       <Buttons
         options={options}
-        onClick={setVisibleOption}
+        onClick={(event, option) => {
+          // @ts-expect-error this is fine
```
MattIPv4: Can we avoid this expect-error?

src/pages/overlay/hooks/useSleeping.tsx
```diff
@@ -63,12 +65,28 @@ export const SleepingProvider = ({ children }: { children: ReactNode }) => {
     stopTimer();
   }, [callbacks.sleep, stopTimer]);
 
+  // Pause the timer and keep the overlay awake
+  const stayAwake = useCallback(() => {
+    setSleeping(false);
+    stopTimer();
+  }, [stopTimer]);
```
MattIPv4: Can we call this caffeinate?

MattIPv4: Should this be calling the wake callbacks?

src/pages/overlay/hooks/useSleeping.tsx
```diff
@@ -63,12 +65,28 @@ export const SleepingProvider = ({ children }: { children: ReactNode }) => {
...
+
+  // Resume the timer
+  const resumeTimer = useCallback(
+    (time: number) => {
+      startTimer(() => setSleeping(true), time);
+    },
+    [startTimer],
+  );
```
MattIPv4, line 108: Can we call this uncaffeinate?

src/pages/overlay/hooks/useSleeping.tsx
MattIPv4: Should this block calls to wake/sleep, which could cause it to sleep, while stay awake is active?

MattIPv4 (commented):
Sorry for taking so long to get to this

src/pages/overlay/components/overlay/Overlay.tsx
flakey5: Yeah, it's supposed to focus the active overlay that was just opened like how it does when setting the active ambassador

MattIPv4: `src/pages/overlay/components/overlay/Ambassadors.tsx` doesn't take a ref, and doesn't have a tabIndex, so that might be why?

(nor does `src/pages/overlay/components/overlay/Settings.tsx`, and `src/components/Welcome.tsx` has a ref but not tabIndex)

flakey5: The root element in `Ambassadors.tsx` doesn't take in a ref, but the parent element of the `AmbassadorButton` does and that's what it's trying to focus

Interestingly adding a `setTimeout` around the `node?.focus()` does cause the element to be focused properly, so I think it's something similar to https://stackoverflow.com/a/76033876 where it's just not triggering a re-render

PR #383, opened 2025-10-27
Add sorting options for ambassadors

### LINKED ISSUE #382: Add sorting options for Ambassadors
Currently, the Ambassadors lists are sorted by the date each ambassador arrived. It would be helpful to add additional sorting options:
- Alphabetical by name
- Alphabetical by species

These options could be exposed through a simple dropdown in the settings UI, allowing users to choose their preferred sorting method.

Suggestion from @just_some_donkus in chat!

I'll take a stab at this later tonight, just wanted to document this so I don't forget!

CtrlNate:
Resolves #382

Currently, the Ambassadors lists are sorted by the date each ambassador arrived. I have added additional sorting options:

- (NEW DEFAULT) Sort in the order they appear in @alveusgg/data
- Arrival date
- Alphabetical by name
- Alphabetical by species

Users can select their default sorting method in the settings menu.
image

### DIFF
--- a/src/pages/overlay/components/Select.tsx
+++ b/src/pages/overlay/components/Select.tsx
@@ -1,6 +1,6 @@
 import { useCallback, type ChangeEvent } from "react";
 
-interface SelectOption<T extends string = string> {
+export interface SelectOption<T extends string = string> {
   value: T;
   label: string;
 }
--- a/src/pages/overlay/hooks/useSettings.tsx
+++ b/src/pages/overlay/hooks/useSettings.tsx
@@ -16,6 +16,7 @@ import {
   sortMethods,
   sortOptions,
   type SortMethod,
+  type SortOption,
 } from "../../../utils/sorting";
 import {
   isValidOverlayKey,
@@ -27,7 +28,7 @@ type Setting = { title: string; configurable: boolean } & (
   | { type: "string"; process: (value: any) => string }
   | {
       type: "select";
-      options: { value: string; label: string }[];
+      options: SortOption[];
       process: (value: any) => string;
     }
 );
--- a/src/utils/sorting.ts
+++ b/src/utils/sorting.ts
@@ -1,4 +1,6 @@
-type SortOption = { value: string; label: string };
+import type { SelectOption } from "../pages/overlay/components/Select";
+
+export type SortOption = SelectOption;
 
 export const sortOptions = [
   { value: "default", label: "Default" },


src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -33,6 +33,21 @@ const settings = {
...
   },
+  ambassadorSort: {
+    title: "Sort ambassadors by",
+    type: "select",
+    options: [
+      { value: "default", label: "Default" },
+      { value: "arrival", label: "Arrival date" },
+      { value: "name", label: "Name (A-Z)" },
+      { value: "species", label: "Species (A-Z)" },
+    ],
```
MattIPv4: Can we move this options stanza into a new `utils/sorting.ts` file as `sortOptions`, and then use that to generate a `sortMethods` array + a `SortMethod` type? So that we're defining these once instead of repeatedly throughout this PR?

src/pages/overlay/components/overlay/Settings.tsx
```diff
@@ -30,6 +31,16 @@ export default function Settings(props: OverlayOptionProps) {
...
+              {setting.type === "select" && "options" in setting && (
+                <Select
+                  label={setting.title}
+                  value={setting.value as string}
+                  options={
+                    setting.options as { value: string; label: string }[]
+                  }
+                  onChange={setting.change as (value: string) => void}
+                />
+              )}
```
MattIPv4: Maybe its time to add some typings to the `settings` constant (`as const satisfies Record<string, Setting>`), so that you aren't needing to cast values here?

src/pages/overlay/components/Select.tsx
```diff
@@ -0,0 +1,43 @@
...
+  label: string;
+  value: T;
+  options: SelectOption<T>[];
+  onChange: (value: T) => void;
+}
+
+export default function Select<T extends string = string>(
+  props: SelectProps<T>,
+) {
+  const { label, value, options, onChange } = props as SelectProps<T>;
```
MattIPv4: Why the type-cast here?

src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -47,7 +63,7 @@ const settings = {
     },
     configurable: false,
   },
-};
+} as const;
```
MattIPv4:
```suggestion
} as const satisfies Record<string, Setting>;
```

and then something like:

```ts
type Setting = { title: string; configurable: boolean } & (
  | { type: "boolean"; process: (value: any) => boolean }
  | { type: "string"; process: (value: any) => string }
  | { type: "select"; options: { value: strig; label: string }[]; process: (value: any) => string }
);
```

src/pages/overlay/components/overlay/Settings.tsx
```diff
@@ -30,6 +31,14 @@ export default function Settings(props: OverlayOptionProps) {
                   onChange={setting.change as (value: boolean) => void}
                 />
               )}
+              {setting.type === "select" && "options" in setting && (
```
MattIPv4:
```suggestion
              {setting.type === "select" && (
```
Once you've added the type as below, I think you shouldn't need this extra check here?

src/utils/helpers.ts
```diff
@@ -9,3 +11,49 @@ export const typeSafeObjectFromEntries = <T extends [string, unknown][]>(
...
+      return [...ambassadors].sort(([, a], [, b]) =>
+        a.species.name.localeCompare(b.species.name, undefined, {
+          sensitivity: "base",
+        }),
+      );
+    case "default":
+    default:
+      return ambassadors;
+  }
+};
```
MattIPv4: I think this should probably be in the sorting util too?

src/utils/sorting.ts
```diff
@@ -0,0 +1,17 @@
...
+  "name",
+  "species",
+];
+
+export const sortOptions: { value: SortMethod; label: string }[] = [
+  { value: "default", label: "Default" },
+  { value: "arrival", label: "Arrival date" },
+  { value: "name", label: "Name (A-Z)" },
+  { value: "species", label: "Species (A-Z)" },
+];
```
MattIPv4:
```suggestion
type SortOption = { value: string; label: string };

export const sortOptions = [
  { value: "default", label: "Default" },
  { value: "arrival", label: "Arrival date" },
  { value: "name", label: "Name (A-Z)" },
  { value: "species", label: "Species (A-Z)" },
] as const satisfies SortOption[];

export type SortMethod = (typeof sortOptions)[number]["value"];
export const sortMethods: SortMethod[] = sortOptions.map(opt => opt.value);
```
I think we can define the options just once and derive everything else from that?

src/pages/overlay/components/Select.tsx
```diff
@@ -0,0 +1,43 @@
+import { useCallback, type ChangeEvent } from "react";
+
+interface SelectOption<T extends string = string> {
+  value: T;
+  label: string;
```
MattIPv4: Can we move this into the sorting util, replacing the current `SortOption` type?

src/pages/overlay/hooks/useSettings.tsx
```diff
@@ -12,11 +12,26 @@ import {
...
   isValidOverlayKey,
   type OverlayKey,
 } from "../components/overlay/Overlay";
 
+type Setting = { title: string; configurable: boolean } & (
+  | { type: "boolean"; process: (value: any) => boolean }
+  | { type: "string"; process: (value: any) => string }
+  | {
+      type: "select";
+      options: { value: string; label: string }[];
```
MattIPv4: Can we use the SortOption type here?

CtrlNate: oops yep

src/pages/overlay/components/Select.tsx
CtrlNate: I kept this here since it's generic to the Select component in case there's any other settings that would need the component - but I can move it if you want!

CtrlNate: Because I think I'd end up importing it from the sorting utils and having
```ts
interface SelectProps<T extends string = string> {
  label: string;
  value: T;
  options: SortOption<T>[];
  onChange: (value: T) => void;
}
```
in the generic select component, which feels less clear, unless i'm losing it?

MattIPv4: I think it'd be fine to move it and use it as the definition for `SortOption`, as it's basically identical (with the added bonus of the generic)

MattIPv4: I suppose you could import it the other way around too, or maybe best to just keep two definitions?

MattIPv4: Nvm I chewed on this more, two definitions is fine, they serve different purposes.

ACA30:
Let me know if this makes more sense, I can also use SelectOption directly but this feels more readable

PR #421, opened 2026-01-15
Update data (Not Applicable + Unknown as lifespan)

Niall-Sharma:
An [issue](https://github.com/alveusgg/data/issues/142) in the data repository talks about replacing `unknown` wild lifespans with `not applicable` to better align with Alveus' messaging. This PR replaces the use of `unknown` with `not applicable` when a wild lifespan is not given by the database.

Closes [issue](https://github.com/alveusgg/data/issues/142)

Let me know if this change is unneeded in this repository, it just felt more useful if a change was made here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "types": "tsc --noEmit"
   },
   "dependencies": {
-    "@alveusgg/data": "0.70.0",
+    "@alveusgg/data": "0.71.0",
     "@headlessui/react": "^2.2.9",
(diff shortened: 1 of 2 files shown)


Niall-Sharma:
Looking into it, I see a lot of different ambassadors are put as unknown than I first believed.  It makes more sense for the data to send the state instead of the extension automatically choosing unknown if a number isn't provided. Back to the drawing board...

Issue #480, opened 2026-04-25
3d effect causing card overflow


MattIPv4:
When the 3d effects are enabled, and the card is at the full height available in the player, hovering over the card can cause the card to overflow beyond the available space leading to the links at the bottom being unclickable.

<img width="634" height="740" alt="Image" src="https://github.com/user-attachments/assets/11e70f6f-dfde-4e4a-8baa-c3583de00e15" />

We likely should fix this in two ways, reducing how extreme the 3d effect is, and adding some top/bottom margin to the cards on hover when 3d effects are enabled, though I am very much open to other thoughts on how best to address this.

MattIPv4:
#511

PR #511, opened 2026-05-27
Add bottom margin to TiltCard when card effects are enabled

itsjayway:
I'm having trouble getting obs or the ffmpeg script to kick off a livestream, but this is the place to add mb-x to add space above where the progress bar would be. #480 

Commenting/uncommenting the added line: 

https://github.com/user-attachments/assets/fa30e913-de1f-4e08-9dd7-0c5f3e91ddd0


MattIPv4:
🤔 `mb-10` seems like quite a lot of space to be adding given how little the card actually expands when hovering right at the bottom. I think we could probably get away with `mb-4` having tested locally?

Thoughts on having that margin only applied on hover, with a transition, so that the card has more space when the effecting isn't happening?

MattIPv4 (approved):

